PR #12, opened 2024-11-01
EIM-57 Instalation UI improved for cases of installing multiple versions

Hahihula:


Fabricio-ESP (approved):
Successfully able to install more than one version of IDF using the installation manager.
On Ubuntu, the checkboxes does not render well yet, this will come later.

src/components/wizard_steps/InstalationProgress.vue
```diff
@@ -54,18 +66,25 @@ export default {
...
       return false;
     },
     startListening: async function () {
       console.log('Listening for tools messages...');
-      this.unlisten = await listen('tools-message', (event) => {
+      this.unlistenTools = await listen('tools-message', (event) => {
         switch (event.payload.action) {
           case 'start':
-            this.tools[event.payload.tool] = {
+            this.tools[this.curently_installing_version][event.payload.tool] = {
```
brianignacio5, line 87: While unlikely it would be good to add validation here for `this.curently_installing_version` which need state == "started" otherwise it will be undefined to avoid undefined key error here.

Also typo `this.currently_installing_version`

PR #16, opened 2024-11-20
Eim 46 styling gui

Hahihula:
Changed the visual appearance of the installer.
image
image
image
image
image
image
image
image
image
image
image
image

### DIFF
--- a/src/components/LoadConfig.vue
+++ b/src/components/LoadConfig.vue
@@ -11,12 +11,13 @@
             <h2 class="section-title">Load Configuration</h2>
             <p class="section-description">Drag & drop TOML file or click to load existing configuration</p>
           </div>
+          <div v-if="config_loaded" class="config-status">
+            ✓ Config loaded
+          </div>
           <n-button @click="load_config" type="error" ghost class="action-button">
             Load Config
           </n-button>
-          <div v-if="Object.keys(rust_settings).length > 0" class="config-status">
-            ✓ Config loaded
-          </div>
+
         </div>
       </div>
     </div>
@@ -79,7 +80,8 @@ export default {
   data: () => ({
     rust_settings: {},
     unlisten_drag_drop: undefined,
-    isDragging: false
+    isDragging: false,
+    config_loaded: false
   }),
   methods: {
     startWizard() {
@@ -104,6 +106,7 @@ export default {
       });
       const _ = await invoke("load_settings", { path: file });
       this.gs();
+      this.config_loaded = true;
     }
   },
   mounted: async function () {
@@ -120,6 +123,7 @@ export default {
         if (file && file.endsWith('.toml')) {
           const _ = await invoke("load_settings", { path: file });
           this.gs();
+          this.config_loaded = true;
         } else {
           emit('user-message', {
             type: 'error',
--- a/src/components/LoadConfig.vue
+++ b/src/components/LoadConfig.vue
@@ -166,7 +166,7 @@ export default {
 }
 
 .dragging .drop-zone {
-  border-color: #dc2626;
+  border-color: #e7352c;
   background-color: #fee2e2;
 }
 
@@ -203,7 +203,7 @@ export default {
 .icon {
   width: 32px;
   height: 32px;
-  color: #dc2626;
+  color: #e7352c;
 }
 
 .option-title {
--- a/src/components/LoadConfig.vue
+++ b/src/components/LoadConfig.vue
@@ -1,67 +1,81 @@
 <template>
   <div class="load-config">
-    <h1 class="title">Please select starting point!</h1>
-    <n-split direction="horizontal" class="content-split">
-      <template #1>
-        <div class="option-panel">
-          <div class="option-content" @dragover.prevent="handleDragOver" @dragleave.prevent="handleDragLeave"
-            @drop.prevent="handleDrop" :class="{ 'dragging': isDragging }">
-            <div class="drop-zone">
-              <div class="icon-container">
-                <svg class="icon" viewBox="0 0 24 24" fill="none" stroke="currentColor">
-                  <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2"
-                    d="M5 19a2 2 0 01-2-2V7a2 2 0 012-2h4l2 2h4a2 2 0 012 2v1M5 19h14a2 2 0 002-2v-5a2 2 0 00-2-2H9a2 2 0 00-2 2v5a2 2 0 01-2 2z" />
-                </svg>
-              </div>
-              <h2 class="option-title">Load Configuration</h2>
-              <p class="option-description">Load an existing configuration file to start the installation process</p>
-              <p class="drag-drop-text">
-                <span class="drag-icon">📄</span>
-                Drag and drop your TOML file here
-                <br>
-                <span class="drag-drop-or">or</span>
-              </p>
-              <n-button @click="load_config" type="error" ghost class="action-button">
-                Load Installation Config
-              </n-button>
-            </div>
-            <div v-if="Object.keys(rust_settings).length > 0" class="config-preview">
-              <h3 class="preview-title">Current Configuration:</h3>
-              <pre>{{ JSON.stringify(rust_settings, null, 2) }}</pre>
-            </div>
+    <h1 class="title">Installation Setup</h1>
+
+    <!-- Compact Config Loading Section -->
+    <div class="config-section" @dragover.prevent="handleDragOver" @dragleave.prevent="handleDragLeave"
+      @drop.prevent="handleDrop" :class="{ 'dragging': isDragging }">
+      <div class="config-content">
+        <div class="config-layout">
+          <div class="config-text">
+            <h2 class="section-title">Load Configuration</h2>
+            <p class="section-description">Drag & drop TOML file or click to load existing configuration</p>
+          </div>
+          <n-button @click="load_config" type="error" ghost class="action-button">
+            Load Config
+          </n-button>
+          <div v-if="Object.keys(rust_settings).length > 0" class="config-status">
+            ✓ Config loaded
           </div>
         </div>
-      </template>
-      <template #2>
-        <div class="option-panel">
-          <div class="option-content">
-            <div class="icon-container">
-              <svg class="icon" viewBox="0 0 24 24" fill="none" stroke="currentColor">
-                <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2"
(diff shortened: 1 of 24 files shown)


src/components/LoadConfig.vue
```diff
@@ -1,23 +1,43 @@
...
+        <div class="option-panel">
+          <div class="option-content">
+            <div class="icon-container">
+              <svg class="icon" viewBox="0 0 24 24" fill="none" stroke="currentColor">
+                <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2"
+                  d="M5 19a2 2 0 01-2-2V7a2 2 0 012-2h4l2 2h4a2 2 0 012 2v1M5 19h14a2 2 0 002-2v-5a2 2 0 00-2-2H9a2 2 0 00-2 2v5a2 2 0 01-2 2z" />
+              </svg>
+            </div>
+            <h2 class="option-title">Load Configuration</h2>
+            <p class="option-description">Load an existing configuration file to start the installation process</p>
```
brianignacio5: Is it possible to implement some drag and drop functionality ? It would be cool to drag and drop the config here to load config files for users. If possible we can add it for future feature request.

src/components/wizard_steps/InstallationPathSelect.vue
```diff
@@ -1,37 +1,84 @@
...
+    <h1 class="title">Select Installation Location</h1>
+
+    <n-card class="path-card">
+      <div class="card-content">
+        <div class="path-info">
+          <h3 class="info-title">ESP-IDF Installation Directory</h3>
+          <p class="info-desc">Choose where to install ESP-IDF and its tools. Ensure you have sufficient disk space.</p>
+        </div>
+
+        <!-- <div class="space-required">
```
brianignacio5: Can remove commented code.

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -104,9 +115,111 @@ export default {
       }
     }
   },
+  computed: {
+    canProceed() {
+      return this.selected_idf_mirror && this.selected_tools_mirror &&
```
brianignacio5, line 120: `this.selected_idf_mirror` used twice

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -1,71 +1,79 @@
...
     selected_tools_mirror: null,
     idf_mirrors: [],
     tools_mirrors: [],
+    defaultMirrors: {
+      idf: '',
+      tools: ''
+    }
   }),
   methods: {
     get_avalible_idf_mirrors: async function () {
```
brianignacio5: `get_available_idf_mirrors `

src/components/wizard_steps/VersionSelect.vue
```diff
@@ -1,37 +1,66 @@
...
+              <div class="version-content">
+                <div class="version-header">
+                  <span class="version-name">{{ version.name }}</span>
+                  <n-tag v-if="version.latest" type="error" size="small">Latest</n-tag>
+                  <n-tag v-if="version.lts" type="success" size="small">LTS</n-tag>
+                </div>
+                <span v-if="version.description" class="version-description">
+                  {{ version.description }}
+                </span>
+                <!-- <div class="version-meta">
```
brianignacio5: Another commented code here.

brianignacio5 (approved):
A few changes but otherwise LGTM

brianignacio5:
Maybe unrelated to this PR but worth looking into it. I've received this error while cmake extraction.

image

kolipakakondal:
Hi @HahihulaDo we detect prerequisites from Homebrew as well? I see CMake is not being detected through its installation.

Screenshot 2024-11-21 at 11 19 21 AM

kolipakakondal:
Can we simplify the layout to accommodate more information in less space? Instead of showing the missing prerequisites below, we could list all prerequisites in the top section itself. After the check, it can display what’s missing and what will be installed using status icons. With that approach, the user can avoid scrolling down to see the Missing Prerequisites section.

Screenshot 2024-11-21 at 11 21 52 AM

kolipakakondal:
For any OS, if users need to install any prerequisites themselves, please provide a link within the wizard. This will help avoid users having to go through the documentation to figure out what needs to be installed.
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/linux-macos-setup.html?highlight=dfu%20util#step-1-install-prerequisites


Screenshot 2024-11-21 at 11 29 10 AM

kolipakakondal:
**Good to have:**
If there is a persistent issue with installing tools or prerequisites and the user would like to check the logs, it would be great if we could provide a link that directly opens the log file.

kolipakakondal:
Hi @Hahihula Thanks for PR. Good progress. See if we can implement the following type of UI: The header, footer, and wizard size should remain fixed across platforms, while the middle section will be used for the installation content.Changing the content area size for every screen disrupts a professional appearance. See if we can consider this approach.

@pedrominatel @igrr Please share your thoughts with us before Petr considers the revamp.

e4f9f72086bf4de4a4b333e3a0033931

src/components/wizard_steps/MirrorSelect.vue
Hahihula: it's IDF mirrot and TOOLS mirror

Hahihula:
> Hi @HahihulaDo we detect prerequisites from Homebrew as well? I see CMake is not being detected through its installation.
> 
Yes, we detect the prerequisites from homebrew. Which version did you used? Isn't this the old issue with the version from dmg package? Only working is probably the binary one. The version with dmg or msi will be used After we start with version management and as they are no use as of today, I did not spend time resolving these permissions issues.

src/components/wizard_steps/MirrorSelect.vue
brianignacio5: xD my mistake.

src/components/LoadConfig.vue
Hahihula: I've added drag&drop ( you can drop the file anywhere on the window in this step ) even with animated highlight .
image
image

src/components/wizard_steps/MirrorSelect.vue
Hahihula: done

src/components/wizard_steps/VersionSelect.vue
Hahihula: done

src/components/wizard_steps/PrerequisitiesCheck.vue
```diff
@@ -1,40 +1,64 @@
...
-          </template>
+  <div class="prerequisites">
+    <h1 class="title">Prerequisites Check</h1>
+    <p class="description">The installer will now verify required components for ESP-IDF...</p>
+
+    <div class="check-section">
+      <n-card class="prerequisites-list">
+        <template #header>
+          <div class="card-header">
+            <span class="header-title">Required Prerequisites</span>
```
pedrominatel: This is redundant. Change just to `Prerequisites`.

src/components/wizard_steps/InstalationProgress.vue
```diff
@@ -1,59 +1,101 @@
...
-                {{ tool.finished ? 'yes' : 'no' }}
-              </td>
-              <td>
-                {{ tool.error ? 'yes' : 'no' }}
-              </td>
-            </tr>
-          </tbody>
-        </table>
+  <div class="installation-progress">
+    <h1 class="title">ESP-IDF Installation Progress</h1>
```
pedrominatel: We can avoid using ESP-IDF everywhere. The installation manager title in red makes this clear, and we could remove it where it is not needed.

src/components/Welcome.vue
```diff
@@ -14,20 +44,100 @@ export default {
...
-.welcome {
+.installer {
+  min-height: 100vh;
+  display: flex;
+  flex-direction: column;
+  background-color: #f5f5f5;
+}
+
+.header {
+  background-color: #dc2626;
```
pedrominatel, line 64: If red is the color, change to the Espressif logo red (#e7352c). I know it's nitpicking but it's part of the branding.

Hahihula: It was meant to be Espressif red, replacing all around the app!

pedrominatel:
@kondal, IMHO we need to keep the UI as simple as possible, with the possibility of installing with some of the configurations settled as default.
I'm attaching this UI for you to take a look at (be kind, I'm not a designer ).

Screenshot 2024-11-21 at 20 33 39

In this proposal, we can simplify the interface by adding the links for each step in a way that the user can return to that specific part or go directly to the installation without setting anything (if possible).
The part with the ESP-IDF version selection and the SoCs can be done in just one step and the version should reflect on the supported target list.

Do we need to show the dependency check results every time? Does this information matter if all dependencies match?

The download server settings can be included in the general setting and should be used only if necessary.

Fabricio-ESP:
Hi @Hahihula ,
I noticed the height of the template window is too high from start, most people still use fullhd screens and the application is occupying all vertical height and still the footer is not visible on startup, forcing users to maximize the window to see it. Even with that the "Select your starting point!" para shows a scroll bar, it is better to make it wider and reduce height.

I guess there are several suggestions to change the main template, just take window height in consideration, and attempt to keep it under 900pixels

Hahihula:
@kolipakakondal  @pedrominatel  Ok, so what about letting the user on the starting page to a)load config b) select "expert mode" instalation (which will lead to current wizard) c) select "default/simplified installation" which will not ask him any question unless hitting a wall ?

pedrominatel:
> @kolipakakondal @pedrominatel Ok, so what about letting the user on the starting page to a)load config b) select "expert mode" instalation (which will lead to current wizard) c) select "default/simplified installation" which will not ask him any question unless hitting a wall ? As we support complete headless and non-interactive installation, this should actually be achievable without any more effort (just with the small UI redo)

Sounds good to me!

Hahihula:
hope this will be ok:
image

kolipakakondal:
> @kolipakakondal, IMHO we need to keep the UI as simple as possible, with the possibility of installing with some of the configurations settled as default. I'm attaching this UI for you to take a look at (be kind, I'm not a designer ).
> 
> Screenshot 2024-11-21 at 20 33 39
> In this proposal, we can simplify the interface by adding the links for each step in a way that the user can return to that specific part or go directly to the installation without setting anything (if possible). The part with the ESP-IDF version selection and the SoCs can be done in just one step and the version should reflect on the supported target list.
> 
> Do we need to show the dependency check results every time? Does this information matter if all dependencies match?
> 
> The download server settings can be included in the general setting and should be used only if necessary.

Hi @pedrominatel  Thanks for the proposal. This looks good to me as well. On the left-hand side, we can show all the steps involved in the installation process and highlight the current one. We will continue to drive the major actions through the bottom button.

I don’t think we need to support back-and-forth navigation to avoid potential issues with the installation process.

In the case of Simplified Installation, we can run all the steps in automatic mode and show only the progress. For Expert Installation, we can allow the user to configure each step and proceed to the next step only by clicking the 'Next' button.

Hahihula:
@kolipakakondal 
image
Here I added the sidebar for the expert installation with visible progress. Not sure if I should add the sidebar elsewhere too as the steps are relevant only for the expert mode.

Hahihula:
Ok, I've added the sidebar to the expert installation:
image

PR #30, opened 2024-12-18
adding data id

Hahihula:


Fabricio-ESP (approved):
With this change it is possible to easily select elements from the app interface.
Auto test already relying on this PR, waiting for it to merge.

PR #31, opened 2025-01-10
test pipeline update

Hahihula:
Fixed wrong platform identification for Intel MacOs. Also, the workflow should be less repeatable now. And for non-release builds the run_id was added for easy identification of the binaries.


kolipakakondal (approved):
LGTM

PR #33, opened 2025-01-15
UX-revamp

Hahihula:


Hahihula:
Header revamp:
image

Hahihula:
Welcome page revamp:
image

Hahihula:
load config revamp:
image

Hahihula:
wizard step redesign:
image

Hahihula:
target select :
image

Hahihula:
version selection:
image

Hahihula:
mirror select:
image

Hahihula:
Path selection:
image

Hahihula:
instalation progress:
image

Hahihula:
finish instalation:
image

Hahihula:
instalation progress bar 
image

Hahihula:
simplified install error message :
image

Hahihula:
added sucesfull progress to simplified install:
image

kolipakakondal:
Typo error "installation" repeated twice - Reuse the configuration from the previous installation

Screenshot 2025-01-22 at 12 06 27 PM

kolipakakondal:
I have to scroll down to see the entire text. Can we avoid this? Perhaps we could reduce the gap between the 'Prerequisites Check' and the 'Expert Installation Step' border.

Screenshot 2025-01-22 at 12 14 36 PM
Screenshot 2025-01-22 at 12 16 21 PM

**Here is what is suggested by Design team**
Screenshot 2025-01-22 at 12 19 04 PM

Hahihula:
image

Fabricio-ESP:
I will add a couple of notes here, it might be applicable for a new tickets and new PR not necessarily changed on this version.

When running simplified installation, but the system is missing prerequisites, the screen lists what is missing, but it keep the loading animation, indicating that something is still going on. Better to remove the animation at this error.
image

On the expert mode, for the prerequisites check, I would recommend to not show the list of requirements, just list what is missing, what is already installed does not need to be shown to the user.
Note that the screen showing missing requirements does not have any action button, to either try again, or cancel, or go back.
image

During the expert Installation, if any selection was made incorrect, there is no option to go back, that feature should be added to allow the user to go back until they select to start the installation.
Note that there isn't any other option rather then start installation or close the application, we should have a button to cancel at any time that would go back to the first screen, and an exit/close button in the first screen as well.

During the installation progress in expert mode I do not think we need to show a table with all the tools being installed, I would just indicating which one is installed and a progress 2/11 ... 3/11, similar to what is done for the simplified installation, the only value added is in case of error, but that will abort the installation and the information will be in the install logs. Anyway, if keeping the table, I will remove the error column, it is misleading, does an x on that table indicated that there was an error there? An error and an X is never easy to understand. Just keep the download, extract and finished column and print ERROR instead of check mark if there was any error there. You will need to show the install log as well, similar to what you do with the simplified install, or at least a button to open it in case of error.
image

When all tools installation is complete, what value does this intermediate screen add? If there was an error it is a good idea to open the logs and an retry/exit/cancel button, but in case of success just move to the next.
image

When saving the configuration file, there is no default name added on the window opened, is there a way to set a default name? Like the config_<datetime> ??
Also on mac it is creating the extension .toml automatically, but on Linux the file is being saved without any extension.

During the installation, using simplified install the text is "Installing ESP-IDF ..." while the expert installation shows "Installation Progress", I would recommend to use similar language regardless for both installation process.

For the last screen the expert installation mode does not provide access to the installation log, maybe it will only show in case of failure, but I see value having it available for users.

Hahihula:
> During the expert Installation, if any selection was made incorrect, there is no option to go back, that feature should be added to allow the user to go back until they select to start the installation.
you can click on any previos step until the installation starts

src/components/wizard_steps/PrerequisitiesCheck.vue
```diff
@@ -121,16 +111,23 @@ export default {
       return false;
     },
   },
+  computed: {
+    percentage() {
+      return this.all_prerequisities.length === 0 ? 0 : ((this.all_prerequisities.length - this.missing_prerequisities.length) / this.all_prerequisities.length) * 100;
+    }
```
Fabricio-ESP, line 117: Might need to round the percentage to any decimal points (or none). Right now it is showing weird numbers:
image

Hahihula: fixed, thanks for the catch!

Fabricio-ESP (approved):
Thanks for the update in the interface.

PR #35, opened 2025-01-20
Fixed release windows extension

Hahihula:
on windows the release artifacts are missing the .exe extension, this should fix it


kolipakakondal (approved):
LGTM

PR #39, opened 2025-01-27
EIM-95 added link to product selector to target select of erpert inst…

Hahihula:
image


Fabricio-ESP:
Any reason you are poiting to a comparison page between a few chips than the homepage on the product selector?

PR #40, opened 2025-01-28
EIM-111 path validation added

Hahihula:
image
image

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -503,6 +503,25 @@ fn set_installation_path(app_handle: AppHandle, path: String) -> Result<(), Stri
     );
     Ok(())
 }
+#[tauri::command]
+fn is_path_empty_or_nonexistent(path: String) -> bool {
+  let path = Path::new(&path);
+  
+  // If path doesn't exist, return true
+  if !path.exists() {
+      return true;
+  }
+  
+  // If path exists, check if it's a directory and if it's empty
+  if path.is_dir() {
+      match fs::read_dir(path) {
+          Ok(mut entries) => entries.next().is_none(), // true if directory is empty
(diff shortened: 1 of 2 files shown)


Fabricio-ESP:
For this feature, it does validate the path as I type with proper update of the banner. One question though:
If I want to install more than one IDF version on the same folder, lets say ~/.espressif, the installer will create a subfolder with the IDF version in there, so I would have ~/.espressif/v5.4 and ~/.espressif/v5.3.2, But on this logic when I enter the path ~/.espressif it shows me as the path is not valid, but still it is where I want to install the new IDF version.
I understand that right now we are not blocking the installer and we can proceed anyways, but the logic to indicate the path is not valid might need to adjust since the default folder would show as invalid if the user is running it for the second time to install a different version.

kolipakakondal (changes requested):
Once we address the validation issue mentioned by @Fabricio-ESP, we can proceed with the merge. Otherwise, LGTM.

Hahihula:
> For this feature, it does validate the path as I type with proper update of the banner. One question though: If I want to install more than one IDF version on the same folder, lets say ~/.espressif, the installer will create a subfolder with the IDF version in there, so I would have ~/.espressif/v5.4 and ~/.espressif/v5.3.2, But on this logic when I enter the path ~/.espressif it shows me as the path is not valid, but still it is where I want to install the new IDF version. I understand that right now we are not blocking the installer and we can proceed anyways, but the logic to indicate the path is not valid might need to adjust since the default folder would show as invalid if the user is running it for the second time to install a different version.

Current installer does not support this. So if the gui will not prevent installation to the folder with some IDF version already present. The installati0on will fail.

Hahihula:
image
image

kolipakakondal (approved):
LGTM

PR #41, opened 2025-01-28
Eim 113 logs button

Hahihula:
image

image

after clicking it opens the folder containing logs
image

works on windows too:
image
 tested on linux too


Fabricio-ESP:
I just checked on Windows/Linux and Mac and when clicking on the "Log folder" icon the app navigates back to the welcome page.
If the user start an expert installation and navigate through a couple of steps and click on the log it will return to the welcome page, if they re-enter the expert installation they can continue from where they left off, but that is an annoyance to navigate there again.
If the installation has already started and the user enter the link, they loose track of the installation progress and cant return to the page.

Hahihula:
> I just checked on Windows/Linux and Mac and when clicking on the "Log folder" icon the app navigates back to the welcome page. If the user start an expert installation and navigate through a couple of steps and click on the log it will return to the welcome page, if they re-enter the expert installation they can continue from where they left off, but that is an annoyance to navigate there again. If the installation has already started and the user enter the link, they loose track of the installation progress and cant return to the page.

Good catch! Fixed.

kolipakakondal (approved):
LGTM

PR #44, opened 2025-01-29
EIM-94 added possibility to return to the main page of the installer

Hahihula:


src/App.vue
```diff
@@ -35,6 +36,24 @@ export default {
     NLayoutContent,
     LogLink
   },
+  methods: {
+    goHome() {
+      let step = this.currentStep;
+      if (step < 7) {
```
kolipakakondal, line 42: Consider using a named constant to make it readable

```
const INSTALLATION_STEP = 7;
if (step < INSTALLATION_STEP) {
//do
}
```

kolipakakondal (approved):
LGTM

PR #48, opened 2025-02-04
Improved tools checksums (EIM-109)

Hahihula:
image
image

### DIFF
--- a/src/components/wizard_steps/InstalationProgress.vue
+++ b/src/components/wizard_steps/InstalationProgress.vue
@@ -31,6 +31,7 @@
                 <tr data-id="tools-table-header">
                   <th>Tool</th>
                   <th>Downloaded</th>
+                  <th>SHA</th>
                   <th>Extracted</th>
                   <th>Finished</th>
                   <th>Error</th>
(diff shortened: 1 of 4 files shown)


brianignacio5:
Install on simplified setup fails with extracting error, i see no log file generated:
Screenshot 2025-02-06 181121

When i tested using the expert setup it hangs on IDF Cmake:

Screenshot 2025-02-06 183851

The setup fails to check prerequisites in MacOS M1 (ninja, cmake dfu-util).

src/components/wizard_steps/InstalationProgress.vue
```diff
@@ -31,6 +31,7 @@
                 <tr data-id="tools-table-header">
                   <th>Tool</th>
                   <th>Downloaded</th>
+                  <th>SHA</th>
```
kolipakakondal, line 34: SHA is used internally to verify the downloaded file. Do we really need to show this to the user?

kolipakakondal (approved):

src/components/wizard_steps/InstalationProgress.vue
Hahihula: how do you suggest informing the user when the checksum verification fails after we removed the popup messages?

kolipakakondal: We could consider combining the “Downloaded” status with “SHA” verification for a more concise status update. If the file is downloaded but does not match the checksum, we can mark it as an invalid file and display an error flag.

Exposing too many internal details with different labels might unnecessarily complicate the UI, especially when the user cannot take any specific action based on them.

Think about it.

brianignacio5 (approved):
LGTM

PR #49, opened 2025-02-06
Installer now logs to file on debug level [EIM-128]

Hahihula:


brianignacio5:
Screenshot 2025-02-07 084417

No log file generated when idf-cmake error appear and progress hangs.

PR #56, opened 2025-02-11
added proper error message with back button to the failed expert install

Hahihula:
image


src/components/wizard_steps/InstalationProgress.vue
```diff
@@ -55,7 +56,12 @@
             </n-table>
           </n-tab-pane>
         </n-tabs>
-        <GlobalProgress messagePosition="right" v-if="!instalation_finished" />
+        <GlobalProgress messagePosition="right" v-if="!instalation_finished && !instalation_failed" />
+        <div v-if="instalation_failed" class="error-message" data-id="error-message">
+          <h3 data-id="error-title">Error during installation:</h3>
+          <p data-id="error-message-text">For more information consult the log file.</p>
```
brianignacio5: I would suggest to add a hyperlink to the log file here (maybe use `log file` as hyperlink), so users can easily jump to the log if needed.

Is it possible to show the error here? I understand that looking the log file is more complete solution, but I don't thing most users would go through the whole thing and would lazily complain about this.

I know Logs folder is listed in the bottom. But when an error happens is good to point where to look.

brianignacio5 (approved):
LGTM with few UI comments.

src/components/wizard_steps/InstalationProgress.vue
Hahihula: image
provided more information to the message,

kolipakakondal (approved):

PR #69, opened 2025-04-02
normalized line endings

Hahihula:
all linebreaks should be uniified


brianignacio5 (approved):
LGTM

Activation script output is now splitting variables in Windows with \r\n as expected.

PR #76, opened 2025-04-10
EIM-173 fix hanging installation end on posix system

Hahihula:
This fix bug where on linux the simple installation finishes but hangs and does signal to the user that it has finished.


kolipakakondal (approved):
LGTM

PR #112, opened 2025-05-26
EIM-204 fixed panic on wizard when tools download fails

Hahihula:


Hahihula:
telegram-cloud-photo-size-4-5882048860124204374-x
Now the installer fails properly when the download of the tool file fails

PR #114, opened 2025-05-27
EIM-182 submodules are now installed as default

Hahihula:


Hahihula:
The issue with `Redundant ProgressBar Instance` is not an issue as it's a completely different progress bar.
But the flashing was corrected as suggested.

kolipakakondal (approved):
LGTM

PR #120, opened 2025-06-04
EIM-207 the install command on windows OS now by default installs the…

Hahihula:


kolipakakondal (approved):
LGTM

Fabricio-ESP (approved):
Working as expected.
UI improvement could be set for future task.

Hahihula:
> Working as expected. UI improvement could be set for future task.

We already have a ticket for it so that it will be addressed.

PR #125, opened 2025-06-09
EIM-208 quoted the strings when creating venv, so they can now contain whitespace characters

Hahihula:
works on macos -> installed idf and build hello_world
works on windows -> installed idf and build hello_world

### DIFF
--- a/src-tauri/src/lib/python_utils.rs
+++ b/src-tauri/src/lib/python_utils.rs
@@ -281,18 +281,11 @@ pub fn pip_install_requirements(
     match std::env::consts::OS {
         "windows" => {
             match command_executor::execute_command_with_env(
-                "powershell",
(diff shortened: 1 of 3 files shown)


src-tauri/src/lib/python_utils.rs
```diff
@@ -281,18 +281,11 @@ pub fn pip_install_requirements(
     match std::env::consts::OS {
         "windows" => {
             match command_executor::execute_command_with_env(
-                "powershell",
```
kolipakakondal, line 284: Please check if this was intentional?

Hahihula: Sure it's intentional. We now instead of calling new shell from the one we are currently running are calling directly python binary on known location inside the current shell. Also I tested it manually just to be sure ;-)

PR #126, opened 2025-06-10
EIM-193 tools download refactoring

Hahihula:
After initial refactoring the tools installation was unified between cli and gui and logic moved to library
the tools installations are now contained in a subfolder named after tool version
globally installed tools are not installed again


Fabricio-ESP (approved):
Confirmed successful installation and correct tools.
Tested on Windows11, MacOS15, Ubuntu24

PR #136, opened 2025-06-24
added python bin path to the PATH

Hahihula:
As per request, I am adding the bin/script python folder to the path, because the IDEs call `CMAKE` directly and not using the `idf.py build`.


alirana01 (approved):
LGTM! will test with the upcoming EIM release

brianignacio5 (approved):
LGTM

PR #144, opened 2025-07-03
EIM-230 added the dl page source code

Hahihula:
assorted dl page:
image


kolipakakondal (commented):
After expanding an item, manually collapsing it didn’t work (perhaps by design). When another item was expanded, the previously expanded one collapsed automatically. However, from an end user’s perspective, it feels like the collapse action isn’t working.

The good thing is that you’re detecting the platform and expanding by default, which is helpful for users.

Fabricio-ESP (approved):
Looks good to me.
Agree that a nice improvement would be to allow collapsing all views, instead of only collapsing in case another group is expanded, but not necessary.

PR #146, opened 2025-07-04
EIM-240 when list command does not find any version, it ends with 0 e…

Hahihula:
When list command does not find any version, it ends with 0 exit code and prompt user to install some version.


Fabricio-ESP (approved):
Correct exit code if missing file.
Thanks

PR #147, opened 2025-07-07
Eim 239 fix path handling

Hahihula:
In some cases, when user user `~` or `.` as path it could lead to some errors, now the path should be expanded to absolute on the application boundary, which should make the path handling more resilient.


Fabricio-ESP (approved):
Working well for install path using ~
Tested on Windows11, Linux and MacOS, both CLI non-interactive and GUI

When using . to represent local folder (e.g. eim install -p ./idf), the installation is properly placed in the idf subfolder but the activation script will use "<currentFolder>/./idf"

PR #149, opened 2025-07-07
EIM-245 fix format of the ESP_IDF_VERSION env variable

Hahihula:


Fabricio-ESP (commented):
Version in environmental variable beincg sored as "5.4" or "5.2" as intended.
Need to adjust when importing data from Espressif IDE to remove the patch version.

Hahihula:
> Version in environmental variable beincg sored as "5.4" or "5.2" as intended. Need to adjust when importing data from Espressif IDE to remove the patch version.

We need to discuss this with the Espressif IDE team if they need the patch version or not.

PR #151, opened 2025-07-07
EIM-254 enabled usage of existing IDF directory trough folder selector

Hahihula:


kolipakakondal (approved):
LGTM.

One of the challenges for users is locating the activation script. For example, if a user selects an existing esp-idf directory like /Users/kondalkolipaka/esp/esp-idf and it gets updated with new content, it becomes difficult - especially for CLI users to figure out where the activation script is, since the esp-idf and its tools are often located in different directories in this case.

Once the installation is complete, it would be helpful to clearly show the path where esp-idf was installed, along with the locations of the tools and the activation script.

Not an high priorty thing for now - it would be great if we can consider this for future.

Hahihula:
> LGTM.
> 
> One of the challenges for users is locating the activation script. For example, if a user selects an existing esp-idf directory like /Users/kondalkolipaka/esp/esp-idf and it gets updated with new content, it becomes difficult - especially for CLI users to figure out where the activation script is, since the esp-idf and its tools are often located in different directories in this case.
> 
> Once the installation is complete, it would be helpful to clearly show the path where esp-idf was installed, along with the locations of the tools and the activation script.
> 
> Not an high priorty thing for now - it would be great if we can consider this for future.

I really do not see how this connects to this exact PR.

kolipakakondal:
> > LGTM.
> > One of the challenges for users is locating the activation script. For example, if a user selects an existing esp-idf directory like /Users/kondalkolipaka/esp/esp-idf and it gets updated with new content, it becomes difficult - especially for CLI users to figure out where the activation script is, since the esp-idf and its tools are often located in different directories in this case.
> > Once the installation is complete, it would be helpful to clearly show the path where esp-idf was installed, along with the locations of the tools and the activation script.
> > Not an high priorty thing for now - it would be great if we can consider this for future.
> 
> I really do not see how this connects to this exact PR. Also, especially for CLI users, every installation ends with information:
> 
> ```
> You have successfully installed ESP-IDF
> for using the ESP-IDF tools inside the terminal, you will find activation scripts inside the base install folder
> sourcing the activation script will setup environment in the current terminal session
> ============================================
> to activate the environment, run the following command in your terminal:
>        source "/Users/USERNAME/.espressif/tools/activate_idf_v5.4.2.sh"
> ============================================
> ```
> 
> so it seems clear enough for me where to find that script

I believe one can install using the GUI version and still work in CLI mode.

The point is, it would be better to show where esp-idf and the tools are installed on the final screen. We can track this in seperate JIRA ticket.

PR #152, opened 2025-07-07
EIM-226 introduce versioning to eim_idf.json file

Hahihula:


Fabricio-ESP (approved):
Thanks for adding it.

alirana01 (approved):
LGTM! Need to run the eim for a final confirmation but only adding version in the file should not break the IDE. Opening a corresponding ticket for this in IDE as well to handle this version there

PR #153, opened 2025-07-08
Eim 238 add tools directory to path before tools check

Hahihula:
Improved the tools installation process to prevent unnecessary tools reinstallation.


Fabricio-ESP (approved):
Confirmed default tools folder is checked for pre-existing tools installation before installation (prevent tools for being overwritten).
Noticed that system wide tools (e.g. cmake, ninja) version is detected before checking the version on the tools folder. It might be recommended to have the tools path before the system path to check for tools already installed.

PR #159, opened 2025-07-15
Added QEMU dependencies to the list of mandatory dependencies

Hahihula:


Fabricio-ESP (approved):
Confirmed updated list of pre-requisites for Linux and Mac. Installation can only proceed once they are all installed.
Confirmed QEMU monitor and debug worked after installing these pre-requisites.

PR #163, opened 2025-07-17
EIM-235 import command can now import multiple IDF instalations from tools_set_config.json file

Hahihula:


kolipakakondal (commented):
Thanks for the PR. It was able to import multiple existing ESP-IDF configurations.

However, I noticed that when the same ESP-IDF version was available in multiple locations, it added both.
[tool_set_config.json](https://github.com/user-attachments/files/21345170/tool_set_config.json)


I’m attaching a screenshot for reference.

image

image

Hahihula:
> in multiple locations, it added both.

That's the expected result. Why would it import only some of the IDFs?

PR #169, opened 2025-07-24
remove unnecesary git dependency

Hahihula:
test build... installer does not need git to install IDF, does IDF needs git to run?


Hahihula:
ok, component manager needs git...

PR #170, opened 2025-07-29
EIM-273 list command now displays IDF path

Hahihula:
image


Fabricio-ESP (approved):
Confirmed working on Windows and MacOS

PR #181, opened 2025-08-21
[Test] Refactor test workflow

Fabricio-ESP:
## Description

This PR makes changes only to the github test workflows.
It filters the artifacts downloaded to the test runners to avoid downloading the offline archives.

Additionally the test results and EIM install logs are integrated into a single upload artifact to reduce the number of files for each run.
The result artifacts may be spread into a long path, but they are only needed to check when tests are failing, so it is a good compromise with the reduction of generated artifacts.

## Testing

---

## Checklist

### DIFF
--- a/.github/workflows/test_cli.yml
+++ b/.github/workflows/test_cli.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Download artifacts
         uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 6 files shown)


Hahihula (approved):
LGTM! I like the bundling of the logs, will be easier to find.

PR #182, opened 2025-08-22
Eim 180 offline installer pipelines setup

Hahihula:


Fabricio-ESP (approved):
Looks Good, thanks for the optimization!

PR #188, opened 2025-08-29
Offline installer added to download page

Hahihula:
it looks like this:
image


kolipakakondal:
Hi @Hahihula Thanks for the PR. Looks good to me. Could we have tabs to differentiate between Online Installers and Offline Installers, so that users can easily navigate based on their use case. By default, we focus on the Online Installers and show them.

Either on the top or left side panel whever you think is the best fit for navigation.

Screenshot 2025-09-02 at 11 09 12 AM

PR #192, opened 2025-09-08
Eim 276 offline install in GUI

Hahihula:
Offline installation now reports progress the same way as the other installation paths
image


kolipakakondal (approved):
LGTM

PR #211, opened 2025-09-26
fixed constrains file search in offline mode

Hahihula:
this should fix the path the constrain file is searched for in the offline mode


Fabricio-ESP (approved):
Confirmed working.
Thanks for the change.

PR #213, opened 2025-09-29
[EIM-309] [Test] Add element to bypass file pick dialog

Fabricio-ESP:
## Description

This PR creates an element in the frontend application that can be manipulated to bypass the file picker dialog for offline installer and loading install configuration file.

This change is required to allow the automated tests to exercise these features without the need to manipulate the dialog to select the decired file.

The change was made in such way that the same functions and events used on regular operation are triggered programatically, and only the file path is provided alternatively to the system dialog.

## Testing
The GUI application tested and confirmed no change to the normal usage of hte application.

---

## Checklist

### DIFF
--- a/src/components/BasicInstaller.vue
+++ b/src/components/BasicInstaller.vue
@@ -125,6 +125,7 @@
           <n-button type="primary" size="large" block>
             Browse Archive File
           </n-button>
+          <input id="eim_offline_installation_input" type="hidden" ref="offlineInput" />
         </div>
       </n-card>
 
@@ -145,6 +146,7 @@
           <n-button type="primary" size="large" block>
             Browse Configuration File
           </n-button>
+          <input id="eim_load_config_input" type="hidden" ref="configInput" />
         </div>
       </n-card>
     </div>
@@ -194,6 +196,10 @@ export default {
     const os = ref('unknown')
     const prerequisitesOk = ref(true)
     const missingPrerequisites = ref([])
+
+    // Input elements for external tests
+    const offlineInput = ref(null);
+    const configInput = ref(null);
 
 
     const checkPrerequisites = async () => {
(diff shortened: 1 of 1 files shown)


src/components/BasicInstaller.vue
```diff
@@ -195,6 +197,10 @@ export default {
     const prerequisitesOk = ref(true)
     const missingPrerequisites = ref([])
 
+    // Input elements for external tests
+    const offlineInput = ref(null);
+    const configInput = ref(null);
```
Hahihula: You are defining a default value to null, but then clearing the value to an empty string. I would suggest clearing it to the default value or undefined.

src/components/BasicInstaller.vue
```diff
@@ -145,6 +146,7 @@
           <n-button type="primary" size="large" block>
             Browse Configuration File
           </n-button>
+          <input id="eim_load_config_input" type="hidden" ref="configInput" />
```
Hahihula: Could you please name it more clearly? Like configInputCiTests or something? So it's purpose is clear and it not get tossed away during some future refactoring.

src/components/BasicInstaller.vue
Fabricio-ESP: Good point, yes, reset them to null now.

src/components/BasicInstaller.vue
Fabricio-ESP: Done, also added a comment in the code for future reference.

src/components/BasicInstaller.vue
```diff
@@ -264,13 +270,19 @@ export default {
...
-          multiple: false,
-          filters: [{
-            name: 'Configuration',
-            extensions: ['json', 'toml']
-          }]
-        })
+        let selected = null
+        if((configInput.value?.value ?? '').trim().length > 0){
+          selected = configInput.value;
+          configInput.value = ''
```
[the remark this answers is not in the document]
Fabricio-ESP: Sure ... Was still experimenting with it.

PR #220, opened 2025-10-02
EIM-211: added the auto_clap functionality to generate completion scripts

alirana01:
## Description

Added the Completions command to help generate the inline cli completions for shells.
Implements # ([EIM-211](https://jira.espressif.com:8443/browse/EIM-211))

## Testing
For testing on macos we can run these commands to see the affect of it for the completion feature

```
mkdir -p ~/.zsh/completions
./eim --generate zsh > ~/.zsh/completions/_eim
chmod go-w ~ ~/.zsh ~/.zsh/completions
fpath+=(~/.zsh/completions)
autoload -Uz compinit; compinit
# map for current name; pick one:
compdef _eim eim        # if you run `eim`
# or:
compdef _eim ./eim=eim  # if you run `./eim`

```
---

## Checklist

### DIFF
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -149,7 +149,6 @@ pub async fn run_cli(cli: Cli) -> anyhow::Result<()> {
         Commands::Completions { shell } => {
             let mut cmd = Cli::command();
             let bin_name = env!("CARGO_PKG_NAME");
-            let shell = shell;
             generate(shell, &mut cmd, bin_name, &mut std::io::stdout());
             return Ok(());
         }


src-tauri/Cargo.toml
```diff
@@ -77,34 +111,46 @@ struct_iterable = "0.1.1"
...
+rustpython-vm = { git = "https://github.com/Hahihula/RustPython.git", branch = "test-rust-build", features = [
+    "freeze-stdlib",
+], optional = true }
+rustpython-stdlib = { git = "https://github.com/Hahihula/RustPython.git", branch = "test-rust-build", features = [
+    "ssl-vendor",
+], optional = true }
 os_info = "3.12.0"
-
-
+clap_complete = "4.5.58"
```
Hahihula: could you add this to the cli dependencies and make it optional and install only if cli feature is enabled?

Hahihula:
Also it would be nice if you update the documentation according to the cli change

Hahihula (changes requested):
Thanks for the contribution. Just couple of nitpicks, as the clap_complete is for cli part, it should be optional dependency under the cli feature like the other cli dependencies (you can see that in cargo.toml)
Also if you are adding this functionality, you should also mention it in the documentation.
and last nitpick is that we do not use the 'feature/fix/something' prefix in the PR names, but on the other hands there should be the ticket designation present.
Otherwise it LGTM!

Hahihula (dismissed):
Thanks for the improvements. LGTM!
just please do not forget to squash the commits during the merge

Fabricio-ESP (approved):
Looks Good.

PR #223, opened 2025-10-03
EIM-312 unify UI of installation using config

Hahihula:
image


Fabricio-ESP (approved):
Thanks for adding that action button.

kolipakakondal (approved):
LGTM

PR #226, opened 2025-10-06
EIM-299 removing link to the IDF shell from windows desktop when removing IDF installation

Hahihula:
currently when IDF version is removed using the version manager, the desktop shortcut stays in place abandoned... this should solve it...
not fully tested yet


Fabricio-ESP (approved):
Confirmed shortcut is now removed from desktop once the IDF version is removed.

PR #242, opened 2025-10-23
EIM-328 initial addition to selectable pypi mirror

Hahihula:
CLI argument for pypi mirror was added
we need to add it to the  GUI too
image
image


Fabricio-ESP (approved):
Confirmed working.

PR #255, opened 2025-11-03
EIM-339 offline installer now check fore prerequisities on non-windows system before the installation start

Hahihula:
Offline installer now check fore prerequisities on non-windows system before the installation start

image

EDIT: updated with exact install command:
image

### DIFF
--- a/src/components/OfflineInstaller.vue
+++ b/src/components/OfflineInstaller.vue
@@ -14,12 +14,44 @@
     <n-card v-if="!installationStarted" class="config-card">
       <h2>{{ $t('offlineInstaller.config.title') }}</h2>
       <n-alert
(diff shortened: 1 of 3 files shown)


Fabricio-ESP:
The message box is very narrow, when a scroll bar is added (In case of Linux systems) it blocks the view and make copying the text difficult. Can we increate the height of that field:
image

Also, on Mac Systems, when we open the offline archive the ready to install window is rendered while the pre-requisites are being evaluated, and then the page is re-rendered with the notification. On Linux there is only a gray screen while the re-requisites are evaluated int eh back. Can we add a loading screen there to prevent confusion?

Fabricio-ESP (approved):
Confirmed working.

PR #259, opened 2025-11-03
For the time of being, we are turning off the codesigning of the windows binaries

Hahihula:
After the certificates troubles are solved, we should revert this commit and than proceed as usual.


brianignacio5 (approved):
LGTM

PR #263, opened 2025-11-05
Easy install path checking fixed

Hahihula:


Fabricio-ESP:
The check is working fine on all OS, my concern is the actual error message. It shows that the "path is not empty", Users might interpret it incorrectly and delete everything from the folder what is not the intend. I believe it would be better to show that the IDF folder is not empty like ~/.espressif/v5.5.1, instead of only .espressif.
image

PR #271, opened 2025-11-12
EIM-112: Added the ping check for mirrors

alirana01:
## Description

Ping test added for the mirrors. We first query the HEAD with a GET request and if that fails we try a normal GET to verify latency and connection

## Testing

Tested the installation via cli, gui using both simple and wizard based flows

---

## Checklist

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -144,7 +144,6 @@ log4rs = { version = "1.3.0", optional = true }
 rustpython-vm = { git = "https://github.com/Hahihula/RustPython.git", branch = "test-rust-build", features = ["freeze-stdlib"], optional = true }
 rustpython-stdlib = { git = "https://github.com/Hahihula/RustPython.git", branch = "test-rust-build", features = ["ssl-vendor"], optional = true }
 os_info = "3.12.0"
-url = "2.5.7"
 semver = "1.0.27"
 
 
--- a/src-tauri/src/gui/commands/installation.rs
+++ b/src-tauri/src/gui/commands/installation.rs
@@ -119,152 +119,6 @@ async fn download_idf(
         detail: Some(rust_i18n::t!("gui.installation.download_preparing").to_string()),
         version: Some(version.to_string()),
     });
-
-    let default_mirror = rust_i18n::t!("gui.installation.default_mirror").to_string();
-    let mirror = settings.idf_mirror.as_deref().unwrap_or(&default_mirror);
-    let app_handle_clone = app_handle.clone();
-    let version_clone = version.to_string();
-
-    let handle = thread::spawn(move || {
-        let mut last_percentage = 0;
-        let mut submodule_count = 0;
-        let mut completed_submodules = 0;
-        let mut total_estimated_submodules = 35; // ESP-IDF typically has ~35 submodules
-        let mut main_repo_finished = false;
-        let mut has_submodules = true;
-
-        loop {
-            match rx.recv() {
-                Ok(ProgressMessage::Update(value)) => {
-                    // Main repository clone progress (0-10%)
-                    if value != last_percentage && (value - last_percentage) >= 10 {
-                        last_percentage = value;
-                        emit_installation_event(&app_handle_clone, InstallationProgress {
-                            stage: InstallationStage::Download,
-                            percentage: (value * 10 / 100) as u32, // Main clone: 0-10%
-                            message: rust_i18n::t!("gui.installation.cloning_repository", version = version_clone.clone()).to_string(),
-                            detail: Some(rust_i18n::t!("gui.installation.repository_progress", percentage = value).to_string()),
-                            version: Some(version_clone.clone()),
-                        });
-                    }
-                }
-
-                Ok(ProgressMessage::SubmoduleUpdate((name, value))) => {
-                    has_submodules = true;
-
-                    // Update estimated total if we're seeing more submodules
-                    if submodule_count > total_estimated_submodules {
-                        total_estimated_submodules = submodule_count + 10;
-                    }
-
-                    // Submodules progress: 10-65% (55% total for submodules)
-                    let submodule_base_progress = 10;
-                    let submodule_range = 55;
-                    let current_submodule_progress = (completed_submodules as f32 / total_estimated_submodules as f32) * submodule_range as f32;
-                    let individual_progress = (value as f32 / 100.0) * (submodule_range as f32 / total_estimated_submodules as f32);
-                    let total_progress = submodule_base_progress + current_submodule_progress as u32 + individual_progress as u32;
-
-                    emit_installation_event(&app_handle_clone, InstallationProgress {
-                        stage: InstallationStage::Download,
-                        percentage: total_progress.min(65),
-                        message: rust_i18n::t!("gui.installation.downloading_submodule", name = name.clone()).to_string(),
-                        detail: Some(rust_i18n::t!("gui.installation.submodule_detail",
-                            current = completed_submodules + 1,
-                            total = total_estimated_submodules,
-                            percentage = value).to_string()),
-                        version: Some(version_clone.clone()),
-                    });
-                }
-
-                Ok(ProgressMessage::SubmoduleFinish(name)) => {
-                    has_submodules = true;
-                    completed_submodules += 1;
-                    submodule_count = completed_submodules; // Track actual count
-
-                    // More frequent updates for submodule completion since it's the main work
-                    let effective_total = total_estimated_submodules.max(completed_submodules);
-                    let submodule_progress = 10 + (completed_submodules * 55 / effective_total);
-                    // let submodule_progress = 10 + (completed_submodules * 55 / total_estimated_submodules.max(completed_submodules));
-
-                    let name_display = name.split('/').last().unwrap_or(&name).replace("_", " ");
-                    emit_installation_event(&app_handle_clone, InstallationProgress {
-                        stage: InstallationStage::Download,
-                        percentage: submodule_progress.min(65) as u32,
-                        message: rust_i18n::t!("gui.installation.completed_submodule", name = name_display).to_string(),
-                        detail: Some(rust_i18n::t!("gui.installation.submodule_progress",
-                            completed = completed_submodules,
-                            total = total_estimated_submodules).to_string()),
-                        version: Some(version_clone.clone()),
-                    });
-
-                    emit_log_message(&app_handle_clone, MessageLevel::Info,
-                        rust_i18n::t!("gui.installation.submodule_completed_log",
-                            name = name,
-                            completed = completed_submodules,
-                            total = total_estimated_submodules).to_string());
-                }
-
-                Ok(ProgressMessage::Finish) => {
-                    main_repo_finished = true;
-
-                    // If no submodules were processed, this means we're done with everything
-                    if !has_submodules {
-                        emit_installation_event(&app_handle_clone, InstallationProgress {
-                            stage: InstallationStage::Extract,
-                            percentage: 65,
-                            message: rust_i18n::t!("gui.installation.download_completed_no_submodules").to_string(),
-                            detail: Some(rust_i18n::t!("gui.installation.repository_cloned").to_string()),
-                            version: Some(version_clone.clone()),
-                        });
-                        break;
-                    }
-                    // If submodules were processed, check if we're actually done
-                    else if completed_submodules > 0 {
-                        // We have processed some submodules, likely we're done
-                        emit_installation_event(&app_handle_clone, InstallationProgress {
-                            stage: InstallationStage::Extract,
-                            percentage: 65,
-                            message: rust_i18n::t!("gui.installation.download_completed").to_string(),
-                            detail: Some(rust_i18n::t!("gui.installation.repository_and_submodules", count = completed_submodules).to_string()),
-                            version: Some(version_clone.clone()),
-                        });
-                        break;
-                    }
-                    // If we got Finish but haven't seen submodules yet, just note main repo is done
-                    else {
-                        emit_installation_event(&app_handle_clone, InstallationProgress {
-                            stage: InstallationStage::Download,
-                            percentage: 10,
-                            message: rust_i18n::t!("gui.installation.main_cloned_waiting").to_string(),
-                            detail: Some(rust_i18n::t!("gui.installation.waiting_submodules").to_string()),
-                            version: Some(version_clone.clone()),
-                        });
-                        // Don't break - keep waiting for submodules
-                    }
-                }
-
-                Err(_) => {
-                    // Channel closed - check our state and finish appropriately
-                    if main_repo_finished {
-                        let final_percentage = if has_submodules { 65 } else { 65 };
-                        emit_installation_event(&app_handle_clone, InstallationProgress {
-                            stage: InstallationStage::Extract,
-                            percentage: final_percentage,
-                            message: rust_i18n::t!("gui.installation.download_completed").to_string(),
-                            detail: Some(if has_submodules {
-                                rust_i18n::t!("gui.installation.submodules_processed", count = completed_submodules).to_string()
-                            } else {
-                                rust_i18n::t!("gui.installation.repository_cloned").to_string()
-                            }),
-                            version: Some(version_clone.clone()),
-                        });
-                    }
-                    break;
-                }
-            }
-        }
-    });
-
     
     let is_simple_installation = app_state::is_simple_installation(&app_handle);
     let mirror_to_use = get_mirror_to_use(&app_handle, MirrorType::IDF, settings, is_simple_installation).await;
@@ -281,7 +135,6 @@ async fn download_idf(
     let idf_path_str = idf_path.to_str().unwrap().to_string();
     let repo_stub = settings.repo_stub.clone();
     let version_owned = version.to_string();
-    let idf_mirror = settings.idf_mirror.clone();
     let recurse_submodules = settings.recurse_submodules.unwrap_or_default();
 
     let result = match std::thread::spawn(move || {
@@ -289,7 +142,7 @@ async fn download_idf(
         &idf_path_str,
         repo_stub.as_deref(),
         &version_owned,
-        idf_mirror.as_deref(),
+        Some(&mirror_to_use),
         recurse_submodules,
         tx,
       )
--- a/src-tauri/src/lib/utils.rs
+++ b/src-tauri/src/lib/utils.rs
@@ -6,7 +6,6 @@ use crate::{
     version_manager::get_default_config_path,
 };
 use anyhow::{anyhow, Result, Error};
-use git2::Repository;
 use log::{debug, error, info, warn};
 use rust_search::SearchBuilder;
 use serde::{Deserialize, Serialize};
@@ -23,11 +22,8 @@ use std::{
     cmp::Ordering,
 };
 
-use anyhow::{anyhow, Result, Error};
-use futures::StreamExt;
 use gix::prelude::*;
 use std::sync::atomic::AtomicBool;
-use log::{debug, error, info, warn};
 use regex::Regex;
 use url::Url;
 
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1568,7 +1568,6 @@ dependencies = [
  "fork",
  "fs_extra",
  "gix",
- "futures",
  "idf-env",
  "indicatif",
  "libz-sys",


src-tauri/src/lib/utils.rs
```diff
@@ -751,6 +760,114 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+    let timeout = Duration::from_millis(3000);
+    info!(
+        "Starting mirror latency checks ({} candidates)...",
+        mirrors.len()
+    );
+    let mut mirror_latency_map = HashMap::new();
+
+    for m in mirrors {
+        let url = m.as_str();
+        match measure_url_score(url, timeout).await {
```
Hahihula: with this approach, you can end with some urls measured by speed of response to HEAD request and some responding to GET request. which is not the best comparison strategy. If you are unable to get the head request for some domain, you should fallback to get for all of them.

src-tauri/src/lib/utils.rs
```diff
@@ -751,6 +760,114 @@ fn is_retryable_error(error: &io::Error) -> bool {
     }
 }
 
+/// Returns the base domain (scheme + host + optional port) from a full URL.
```
Hahihula: either this comment or the function is not correct as this implementation toss the port out

src-tauri/src/lib/utils.rs
```diff
@@ -751,6 +760,114 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+/// Return URL -> score (lower is better). Unreachable mirrors get u32::MAX.
+pub async fn calculate_mirror_latency_map(mirrors: &[String]) -> HashMap<String, u32> {
+    let timeout = Duration::from_millis(3000);
+    info!(
+        "Starting mirror latency checks ({} candidates)...",
+        mirrors.len()
+    );
+    let mut mirror_latency_map = HashMap::new();
+
+    for m in mirrors {
```
Hahihula: you are introducing new `futures = "0.3.31"` dependency, why not use it here to check the mirrors in paralel

src-tauri/Cargo.toml
```diff
@@ -82,6 +82,8 @@ once_cell = "1.21.3"
 idf-env = { git = "https://github.com/espressif/idf-env", rev="fd69ab4f550ef35647bb32d1584caa6623cbfc4e" }
 fs_extra = { version = "1.3.0", optional = true }
 lnk = "0.6.3"
+futures = "0.3.31"
```
Hahihula: this added dependency is unused if you do not plan to use it remove it

src-tauri/src/gui/commands/settings.rs
```diff
@@ -220,20 +222,56 @@ pub fn get_idf_mirror_list(app_handle: AppHandle) -> Value {
   };
 
   let mirror = settings.idf_mirror.clone().unwrap_or_default();
-  let mut available_mirrors = idf_im_lib::get_idf_mirrors_list().to_vec();
+  let mut available_mirrors: Vec<String> = idf_im_lib::get_idf_mirrors_list()
```
Hahihula: this is unnecessary allocation, you can work with the &str, no need to copy it over.

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -109,44 +134,122 @@ export default {
...
+      const res = await invoke("get_idf_mirror_urls", {});
+      const urls = res.mirrors || [];
+      const list = urls.map((url) => ({ value: url, label: url, ping: null }));
+      this.idf_mirrors = list;
+      this.selected_idf_mirror = res.selected || (list[0] ? list[0].value : null);
+      this.defaultMirrors.idf = this.selected_idf_mirror || '';
       this.loading_idfs = false;
-      return false;
+      // kick off streaming checks
+      invoke("start_idf_mirror_latency_checks", {}).catch(() => {});
```
Hahihula: what is the point of `.catch(() => {});`

src-tauri/src/lib/settings.rs
```diff
@@ -515,4 +516,40 @@ impl Settings {
...
+    pub async fn get_pypi_mirror_latency_map(&self) -> Result<HashMap<String, u32>> {
+        let available_mirrors = crate::get_pypi_mirrors_list()
+            .to_vec()
+            .iter()
+            .map(|s| s.to_string())
+            .collect::<Vec<String>>();
+        let mirror_latency_map =
+            crate::utils::calculate_mirror_latency_map(&available_mirrors).await;
+        Ok(mirror_latency_map)
+    }
```
alirana01, line 489: @Hahihula I think i made it redundant by adding these functions here or I may have miss understood something

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +758,171 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+        .build()
+        .ok()?;
+
+    // Try HEAD first
+    let start = Instant::now();
+    match client.head(&base_url).send().await {
+        Ok(resp) if resp.status().is_success() => {
+            return Some(start.elapsed().as_millis().min(u32::MAX as u128) as u32);
+        }
+        _ => {
```
Hahihula, line 815: You will need to return Err here so the fallback on the `calculate_mirror_latency_map` level actually work.
now you still have the fallback here

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +758,171 @@ fn is_retryable_error(error: &io::Error) -> bool {
         _ => false,
     }
 }
+/// Returns the base domain from a full URL.
+fn get_base_url(url_str: &str) -> Option<String> {
```
Hahihula, line 779: This function still removes the port information as i was warning you on the previous iteration. You moved it but not fixed it.
use something like:
```
if let Some(port) = url.port() {
      // Port is explicitly specified and non-default
      Some(format!("{}://{}:{}", scheme, host, port))
  } else {
      // No explicit port (or it's the default for the scheme)
      Some(format!("{}://{}", scheme, host))
  }
```

src-tauri/src/lib/utils.rs
```diff
@@ -1,26 +1,34 @@
...
+use log::{debug, error, info, warn};
 use regex::Regex;
+use reqwest::header::{RANGE, USER_AGENT};
+use rust_search::SearchBuilder;
+use serde::{Deserialize, Serialize};
+use tar::Archive;
+use url::Url;
+use zstd::{decode_all, Decoder};
+
+use crate::{
```
Hahihula: Please avoid needles reformatting/rearranging of imports as it masks history

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +758,171 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+    info!(
+        "Starting mirror latency checks ({} candidates)...",
+        mirrors.len()
+    );
+    let mut mirror_latency_map = HashMap::new();
+    let mut head_latency_failed = false;
+
+    for m in mirrors {
+        let url = m.as_str();
+        if !head_latency_failed {
```
Hahihula, line 864: please fix this multiple head_latency_failed checks

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +758,171 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+            (m, score)
+        })
+        .collect();
+
+    entries.sort_by_key(|e| e.1);
+    entries
+}
+
+/// Turn `(url, latency)` tuples into display strings like `https://... (123 ms)` or `(... timeout)`.
+pub fn mirror_entries_to_display(entries: &[MirrorEntry]) -> Vec<String> {
```
Hahihula: as this is formating fo cli part only, it should not be part of common library shared between cli and gui

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +758,171 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+                format!("{u} (timeout)")
+            } else {
+                format!("{u} ({s} ms)")
+            }
+        })
+        .collect()
+}
+
+/// Strip the latency suffix back to a plain URL.
+pub fn url_from_display_line(selected: &str) -> String {
```
Hahihula: if it is really needed it's same issue as above with mirror_entries_to_display ... cli specific code should not be part of the common library.

src-tauri/src/lib/settings.rs
Hahihula: so why do you add them?

src-tauri/src/cli/prompts.rs
```diff
@@ -176,49 +180,100 @@ pub fn check_and_install_python(
...
+    set_value: FSet,         // e.g. |c: &mut Settings, v| c.idf_mirror = Some(v)
+    candidates: Vec<String>, // list of mirror URLs
+    wizard_key: &str,        // e.g. "wizard.idf.mirror"
+    log_prefix: &str,        // e.g. "IDF", "Tools", "PyPI"
+) -> Result<(), String>
+where
+    FGet: Fn(&Settings) -> &Option<String>,
+    FSet: Fn(&mut Settings, String),
+{
+    let interactive = config.non_interactive == Some(false);
```
Hahihula: this is non idiomatic way and also it's reversing the current logic. Curently when non_insteractive is None the wizard work in interactive mode in your case it's the opposite.
just use unwarp_or_default() as all the settings has implemented defaults.

src-tauri/src/gui/commands/installation.rs
```diff
@@ -128,11 +128,11 @@ app_handle: &AppHandle,
                     if value != last_percentage && (value - last_percentage) >= 10 {
                         last_percentage = value;
                         emit_installation_event(&app_handle_clone, InstallationProgress {
-                            stage: InstallationStage::Download,
-                            percentage: (value * 10 / 100) as u32, // Main clone: 0-10%
+                                stage: InstallationStage::Download,
```
Hahihula: avoid formating only changes, again, it's masking the history of the changes and makes the review process unplesent

src-tauri/src/lib/settings.rs
Hahihula, line 489: this 3 functions have multiple flaws... first, why do you use result when it can not fail? 
second:
```
let mirror_latency_map =
            crate::utils::calculate_mirror_latency_map(&available_mirrors).await;
        Ok(mirror_latency_map)
```
should be written as:
```
Ok(crate::utils::calculate_mirror_latency_map(&available_mirrors).await)
```
and ideally without the unnecesary result as this:
```
crate::utils::calculate_mirror_latency_map(&available_mirrors).await
```

src-tauri/src/cli/prompts.rs
```diff
@@ -176,49 +180,100 @@ pub fn check_and_install_python(
...
+            set_value(config, url.clone());
+        }
     }
 
+    Ok(())
+}
+
+pub async fn select_mirrors(mut config: Settings) -> Result<Settings, String> {
+    // IDF mirror
+    let idf_candidates: Vec<String> = idf_im_lib::get_idf_mirrors_list()
```
Hahihula: you are cloning the mirror list just to iterate over it.

src-tauri/src/gui/commands/idf_tools.rs
```diff
@@ -353,6 +366,19 @@ pub async fn setup_tools(
       }
     };
 
+    let default_pypi_mirror = settings.pypi_mirror.as_deref().unwrap();
+    let pypi_mirror_latency_map = settings.get_pypi_mirror_latency_map().await.unwrap();
+    let best_pypi_mirror = pypi_mirror_latency_map
+        .iter()
+        .min_by_key(|(_, latency)| *latency)
+        .unwrap()
```
Hahihula: this will cause panic! ... for example if two keys will have same value min_by_key return None. None does not have `.0` ...
plese awoid using unwrap if you are not 100% sure it's the right thing to do!!

src-tauri/src/gui/commands/installation.rs
```diff
@@ -253,8 +253,23 @@ app_handle: &AppHandle,
...
+    let default_mirror_str = rust_i18n::t!("gui.installation.default_mirror").to_string();
+    let default_mirror = settings
+        .idf_mirror
+        .as_deref()
+        .unwrap_or(&default_mirror_str);
+    let mirror_latency_map = settings.get_idf_mirror_latency_map().await.unwrap();
+    let best_mirror = mirror_latency_map
+        .iter()
+        .min_by_key(|(_, latency)| *latency)
+        .unwrap()
```
Hahihula: this is duplicated logic from the idf_tools. ... it's duplicated error which will cause the app to crash.

src-tauri/src/gui/commands/installation.rs
```diff
@@ -253,8 +253,23 @@ app_handle: &AppHandle,
...
+        .0
+        .clone();
+    let mirror = match best_mirror.is_empty() {
```
Hahihula: here, you select the mirror to be either the best or the default mirror, than emit in in the message and never use it!

src-tauri/src/gui/commands/settings.rs
```diff
@@ -219,21 +216,50 @@ pub fn get_idf_mirror_list(app_handle: AppHandle) -> Value {
       }
   };
 
-  let mirror = settings.idf_mirror.clone().unwrap_or_default();
-  let mut available_mirrors = idf_im_lib::get_idf_mirrors_list().to_vec();
+    let mirror = settings.idf_mirror.clone().unwrap_or_default();
```
Hahihula: another unnecessary formating change... please avoid

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +758,171 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+        Err(e) => {
+            warn!("Mirror ping failed for {}: {:?}", base_url, e);
+        }
+    }
+
+    None
+}
+
+/// Return URL -> score (lower is better). Unreachable mirrors get u32::MAX.
+pub async fn calculate_mirror_latency_map(mirrors: &[String]) -> HashMap<String, u32> {
```
Hahihula: the function as it's written now will introduce potentially multiple 10s of second wait for timeout as it checks the mirrors sequentially, it also creates new reqwest::Client for every mirror. Creating a client is expensive (allocating connection pools, TLS configuration).

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -109,44 +135,96 @@ export default {
...
-          label: mirror,
+    refreshListFromStore(type) {
+      const urls = type === 'idf' ? this.mirrorsStore.idf_urls : type === 'tools' ? this.mirrorsStore.tools_urls : this.mirrorsStore.pypi_urls;
+      const latencyMap = type === 'idf' ? this.mirrorsStore.idf_latency_map : type === 'tools' ? this.mirrorsStore.tools_latency_map : this.mirrorsStore.pypi_latency_map;
+      const selectedFromStore = type === 'idf' ? this.mirrorsStore.selected_idf : type === 'tools' ? this.mirrorsStore.selected_tools : this.mirrorsStore.selected_pypi;
+
+      const list = (urls || []).map((url) => ({
+        value: url,
+        label: url,
+        ping: this.normalizePingValue(latencyMap ? latencyMap[url] : undefined),
```
Hahihula: as this is the only place you use `normalizePingValue` the function is there only for converting hardcoded undefined to hardcoded null...

src-tauri/src/gui/commands/installation.rs
```diff
@@ -253,8 +253,23 @@ app_handle: &AppHandle,
         }
     });
 
-    let default_mirror = rust_i18n::t!("gui.installation.default_mirror").to_string();
-    let mirror = settings.idf_mirror.as_deref().unwrap_or(&default_mirror);
+    let default_mirror_str = rust_i18n::t!("gui.installation.default_mirror").to_string();
+    let default_mirror = settings
```
Hahihula: this is not only default mirror but potentially also user selected preferred mirror

src-tauri/src/lib/settings.rs
```diff
@@ -35,7 +35,7 @@ pub struct Settings {
     pub config_file_save_path: Option<PathBuf>,
     pub non_interactive: Option<bool>,
     pub wizard_all_questions: Option<bool>,
-    pub mirror: Option<String>,
+    pub tools_mirror: Option<String>,
     pub idf_mirror: Option<String>,
```
Hahihula, line 41: This is a breaking API change. Will force to upgrade all integrations. Docker images and actions.

src-tauri/src/cli/cli_args.rs
```diff
@@ -312,7 +312,8 @@ impl IntoIterator for InstallArgs {
                 "tools_json_file".to_string(),
                 self.tools_json_file.map(Into::into),
             ),
-            ("mirror".to_string(), self.mirror.map(Into::into)),
+            // map CLI flag `-m|--mirror` to settings field `tools_mirror`
```
Hahihula: Mapping the CLI --mirror argument to the tools_mirror setting is inconsistent and unjustified. You apply this remapping only for this single argument, which means it behaves differently from all other CLI options. This breaks the established logic for how settings are populated.

Additionally, the CLI is not the only source of configuration. By introducing this one-off mapping, you are decoupling settings in a way that disrupts the overall configuration flow. This is not acceptable.

I already raised this concern in the previous iteration of the review, yet the issue not only remains unresolved, but it has been made worse.

On top of that, the renaming you introduced constitutes a breaking change to the application’s interface. All related changes must be reverted. If such a renaming is ever considered, it must be evaluated and delivered in a dedicated PR, not mixed in with unrelated work.

Do not bundle multiple unrelated changes into the same PR. While it may seem convenient to merge, it significantly increases the complexity and risk when a rollback is needed. My priority is not how easy it is to merge your changes — it’s how difficult it will be to revert them if something breaks. Please follow proper PR scoping and revert the unrelated interface changes before resubmitting.

src-tauri/src/cli/helpers.rs
```diff
@@ -140,3 +144,34 @@ pub async fn track_cli_event(event_name: &str, additional_data: Option<serde_jso
       "additional_data": additional_data
     })).await;
 }
+
+/// Sort mirrors by measured latency (ascending), using None for timeouts.
+pub async fn sorted_mirror_entries(mirrors: &[&str]) -> Vec<MirrorEntry> {
+    let latency_map = calculate_mirror_latency_map(&mirrors.to_vec()).await;
+    let mut entries: Vec<MirrorEntry> = Vec::new();
+    for (key, value) in latency_map.iter() {
+        entries.push((key.clone(), value.clone()));
```
Hahihula: Why are you building a new structure by cloning everything? Please prefer iterator-based construction via collect() for simplicity and zero-cost abstraction. Also you wouldn't need mutable entries

src-tauri/src/cli/helpers.rs
```diff
@@ -140,3 +144,34 @@ pub async fn track_cli_event(event_name: &str, additional_data: Option<serde_jso
       "additional_data": additional_data
     })).await;
 }
+
+/// Sort mirrors by measured latency (ascending), using None for timeouts.
+pub async fn sorted_mirror_entries(mirrors: &[&str]) -> Vec<MirrorEntry> {
+    let latency_map = calculate_mirror_latency_map(&mirrors.to_vec()).await;
```
Hahihula: `&mirrors.to_vec()` converts a slice to a vec and back to a slice reference

src-tauri/src/cli/helpers.rs
```diff
@@ -8,6 +8,10 @@ use std::{
     fmt::Write,
     time::{Duration, Instant},
 };
+use idf_im_lib::utils::{calculate_mirror_latency_map};
+
+/// A tuple containing a mirror URL and its measured latency.
+pub type MirrorEntry = (String, Option<u32>);
```
Hahihula: you should use struct instead of tuple so you are not later accessing it by mystique `e.1`
also use macro to derive Ord to make the sorting cleaner

something like this:
```
impl PartialOrd for MirrorEntry {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        self.latency.partial_cmp(&other.latency)
    }
}

impl Ord for MirrorEntry {
    fn cmp(&self, other: &Self) -> Ordering {
        self.partial_cmp(other).unwrap_or(Ordering::Equal)  // Or handle None explicitly
    }
}
```

src-tauri/src/cli/helpers.rs
```diff
@@ -140,3 +144,34 @@ pub async fn track_cli_event(event_name: &str, additional_data: Option<serde_jso
...
+
+/// Sort mirrors by measured latency (ascending), using None for timeouts.
+pub async fn sorted_mirror_entries(mirrors: &[&str]) -> Vec<MirrorEntry> {
+    let latency_map = calculate_mirror_latency_map(&mirrors.to_vec()).await;
+    let mut entries: Vec<MirrorEntry> = Vec::new();
+    for (key, value) in latency_map.iter() {
+        entries.push((key.clone(), value.clone()));
+    }
+
+    entries.sort_by_key(|e| e.1);
```
Hahihula: This sorting is flawed. It actually sorts timeouts(None) first.

src-tauri/src/gui/commands/idf_tools.rs
```diff
@@ -95,6 +95,7 @@ pub async fn setup_tools(
     offline_archive_dir: Option<&Path>,
 ) -> Result<Vec<String>> {
     info!("Setting up tools...");
+    let is_simple_installation = crate::gui::app_state::is_simple_installation(&app_handle);
```
Hahihula, line 102: is_simple_installation will always evaluate to false in this context. The current logic sets the flag to true in start_simple_setup, but start_simple_setup delegates to start_installation, which then sets it back to false. Since setup_tools is executed after this reset, the flag is never actually true, and the entire conditional flow dependent on is_simple_installation is effectively dead code.

This indicates a fundamental flaw in the implementation and strongly suggests the feature was not tested. Please revisit the logic, correct the control flow, and verify the behaviour before resubmitting.

src-tauri/src/gui/commands/installation.rs
```diff
@@ -254,35 +254,31 @@ app_handle: &AppHandle,
...
-        .iter()
-        .min_by_key(|(_, latency)| *latency)
-        .unwrap()
-        .0
-        .clone();
-    let mirror = match best_mirror.is_empty() {
-        true => default_mirror,
-        false => best_mirror.as_str(),
-    };
+    let is_simple_installation = app_state::is_simple_installation(&app_handle);
```
Hahihula, line 123: as mentioned above, this logic is broken

src-tauri/src/gui/commands/installation.rs
```diff
@@ -988,6 +986,8 @@ pub async fn start_installation(app_handle: AppHandle) -> Result<(), String> {
 /// Starts a simple setup process that automates the installation
 #[tauri::command]
 pub async fn start_simple_setup(app_handle: tauri::AppHandle) -> Result<(), String> {
+    let app_state = app_handle.state::<crate::gui::app_state::AppState>();
+    app_state::set_is_simple_installation(&app_handle, true)?;
```
Hahihula, line 854: as mentioned above this true will always get overwritten by start_simple_setup to false

src-tauri/src/gui/commands/utils_commands.rs
```diff
@@ -13,7 +13,7 @@ use tauri::AppHandle;
 use num_cpus;
 use anyhow::{Result};
 
-use crate::gui;
+use crate::gui::{self, app_state};
```
Hahihula: You added app_state to the import list here, but it is never used. Introducing unused imports is unnecessary noise and should be caught before the PR is submitted.

It would be extremely helpful if you reviewed the GitHub diff yourself before requesting a review. Basic self-review should catch obvious issues like unused imports, leftover code, or incomplete refactoring. This is a matter of professional courtesy and respect for the reviewer’s time.

Please ensure you perform a thorough pass over your changes before assigning the PR for review.

src-tauri/src/gui/utils.rs
```diff
@@ -40,6 +41,28 @@ pub fn is_path_empty_or_nonexistent(path: &str, versions: &[String]) -> bool {
     }
 }
 
+pub async fn get_best_mirror(mirror_latency_map: &HashMap<String, Option<u32>>) -> Option<String> {
```
Hahihula: This String + Option<u32> pairing is fundamentally the same data structure as the tuple you’re using in the CLI component. As mentioned in the previous review round, you should introduce a proper struct to represent a mirror entry and implement Ord (as previously suggested).

If you define this struct in the shared library layer, you can use it consistently in both the CLI and GUI code. This will simplify the implementation, remove the need for ad-hoc comparisons, and ensure that both parts of the application follow the exact same logic when selecting the best mirror.

Please refactor accordingly to avoid duplicating logic and relying on loosely structured tuples or paired values.

src-tauri/src/cli/helpers.rs
Hahihula: much more readable way to write this is 
entries.sort_by_key(|(url, latency)| latency);

src-tauri/src/gui/commands/settings.rs
```diff
@@ -1,19 +1,16 @@
-use tauri::{AppHandle, Manager};
-use idf_im_lib::{settings,to_absolute_path, utils::is_valid_idf_directory};
-use crate::gui::{
-  app_state::{self, get_locked_settings, get_settings_non_blocking, update_settings, AppState},
-  ui::send_message,
-  utils::is_path_empty_or_nonexistent,
-};
+use std::path::PathBuf;
```
Hahihula: Please avoid reformatting imports or modifying code that is unrelated to the scope of your changes. Splitting and rearranging imports without necessity introduces noise, obscures git history, and makes the review process significantly harder.

This is the third time I’ve had to raise this concern on this PR, and it is becoming a recurring issue. Going forward, do not make non-essential formatting or structural changes unless they are directly required for the feature or fix you are implementing.

Thank you for addressing this properly in future submissions.

src-tauri/src/lib/settings.rs
alirana01: reverted back!

src-tauri/src/cli/cli_args.rs
alirana01: reverted

src-tauri/src/cli/helpers.rs
alirana01: removed here

src-tauri/src/cli/helpers.rs
alirana01: removed at required places

src-tauri/src/cli/helpers.rs
alirana01: done

src-tauri/src/cli/helpers.rs
alirana01: fixed with struct impl Ord

src-tauri/src/gui/commands/idf_tools.rs
alirana01: separate command added that will be called from gui avoiding this

src-tauri/src/gui/commands/installation.rs
alirana01: done

src-tauri/src/gui/commands/installation.rs
alirana01: fixed

src-tauri/src/gui/commands/utils_commands.rs
alirana01: removed

src-tauri/src/gui/utils.rs
alirana01: moved original struct to the library now

src-tauri/src/cli/helpers.rs
alirana01: removed here since now it is part of shared library as a struct

src-tauri/src/gui/commands/settings.rs
alirana01: cleaned up

src-tauri/src/lib/utils.rs
```diff
@@ -19,8 +19,40 @@ use std::{
...
+}
+
+impl PartialOrd for MirrorEntry {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for MirrorEntry {
+    fn cmp(&self, other: &Self) -> Ordering {
```
Hahihula, line 43: Much better approach! But still ... Option<T> already implements the Ord Trait. So it can be simplified like this:
```
match (self.latency, other.latency) {
    (Some(a), Some(b)) => a.cmp(&b),
    (Some(_), None) => Ordering::Less,
    (None, Some(_)) => Ordering::Greater,
    (None, None) => Ordering::Equal,
}
```
This 1) avoids unwrap() 2) ensures all possible cases are covered (match will not let you have unimplemented variants) 3) is clearly readable 4) pattern matching is the rust way of doing things ;-)

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +782,127 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+                    info!("Mirror score: {} -> {}", url, score);
+                    mirror_latency_map.insert(url.to_string(), Some(score));
+                }
+                Err(e) => {
+                    warn!("{}", e.to_string());
+                    head_latency_failed = true;
+                }
+            }
+        }
+        if head_latency_failed {
```
Hahihula: you can use `else` here

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +782,127 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+        Err(e) => {
+            warn!("Mirror ping failed for {}: {:?}", base_url, e);
+        }
+    }
+
+    None
+}
+
+/// Return URL -> score (lower is better). Unreachable mirrors get None.
+pub async fn calculate_mirror_latency_map(mirrors: &[&str]) -> HashMap<String, Option<u32>> {
```
Hahihula: Also as you now have the `MirrorEntry` type, with ordering implemented, I would expect you to realize that you can make select_single_mirror and select_best_mirror much better. If you use Vec<MirrorEntry> in the calculate_mirror_latency_map function.
Is more idiomatic and more efficient, it's also more readable as you know you are working with  MirrorEntry also thsi way you will avoind new alocation of the string keys.

src-tauri/src/cli/prompts.rs
```diff
@@ -177,48 +179,100 @@ pub fn check_and_install_python(
...
-        config.mirror = Some(generic_select(
-            "wizard.tools.mirror",
-            &idf_im_lib::get_idf_tools_mirrors_list()
-                .iter()
-                .map(|&s| s.to_string())
-                .collect(),
-        )?)
+    // Only measure mirror latency if we actually need a value (or wizard wants to ask)
+    if interactive && (wizard_all || needs_value) {
+        let mut entries = calculate_mirror_latency_map(candidates).await.into_iter()
```
Hahihula: after you convert the calculate_mirror_latency_map so it returns Vec<MirrorEntry>  you can than
just `entries.sort();` without the unnecessary rearranging everything with unnecessary allocations

src-tauri/src/gui/utils.rs
```diff
@@ -40,6 +42,21 @@ pub fn is_path_empty_or_nonexistent(path: &str, versions: &[String]) -> bool {
     }
 }
 
+pub async fn get_best_mirror(mirror_latency_map: HashMap<String, Option<u32>>) -> Option<String> {
```
Hahihula: This function will also benefit from converting mirror_latency_map to Vec<MirrorEntry>

src/main.js
```diff
@@ -26,6 +27,14 @@ app.use(i18n);
 app.use(createPinia());
 app.use(router);
 app.use(naive);
+
+// Bootstrap background mirror latency on app launch
+try {
```
Hahihula: You are blocking here, loading the whole application GUI for all the time till requests to the mirrors timeout... 
Please initialise the store non-blocking way.  You can already see an example of how to do that in the application code.
Please, you really need to test every new feature before submitting it to PR. You must notice the whitescreen before the application starts blocking the loading of the GUI.

src/store.js
```diff
@@ -427,3 +427,155 @@ export const useWizardStore = defineStore("wizard", {
     },
   },
 });
+
+export const useMirrorsStore = defineStore("mirrors", {
```
Hahihula, line 431: I would suggest to you that you refactor the UI part of the code according to your improvements in the CLI. Mostly the ones you are going to implement in this iteration of review.

src/store.js
```diff
@@ -427,3 +427,155 @@ export const useWizardStore = defineStore("wizard", {
...
+      // Fetch quick URL lists + defaults for all types in parallel
+      this.loading_idf_urls = true;
+      this.loading_tools_urls = true;
+      this.loading_pypi_urls = true;
+      try {
+        const pIdf = invoke("get_idf_mirror_urls", {});
+        const pTools = invoke("get_tools_mirror_urls", {});
+        const pPypi = invoke("get_pypi_mirror_urls", {});
+
+        const [idf, tools, pypi] = await Promise.allSettled([pIdf, pTools, pPypi]);
```
Hahihula: This is blocking! Please avoid blocking the main thread. This is causing the UI to "freeze".

src/store.js
```diff
@@ -427,3 +427,155 @@ export const useWizardStore = defineStore("wizard", {
...
+      // Kick off progressive per-type background latency calculations
+      this.computeLatencyInBackground();
+    },
+
+    computeLatencyInBackground() {
+      const now = Date.now();
+      // IDF
+      if (!this.ttlValid(this.idf_last_updated) && !this.loading_idf_latency) {
+        this.loading_idf_latency = true;
+        invoke("get_idf_mirror_list", {})
```
Hahihula: If you refactor Rust to return Vec<MirrorEntry>, then:

The Tauri commands (get_idf_mirror_list, etc.) should return an array.

src/store.js
```diff
@@ -427,3 +427,155 @@ export const useWizardStore = defineStore("wizard", {
...
+    toolsUrls: (state) => state.tools_urls,
+    pypiUrls: (state) => state.pypi_urls,
+    idfLatencyMap: (state) => state.idf_latency_map,
+    toolsLatencyMap: (state) => state.tools_latency_map,
+    pypiLatencyMap: (state) => state.pypi_latency_map,
+  },
+  actions: {
+    // Backend uses Option<u32> for latency values; Timedout values are represented as None. 
+    // We normalize to 0 for timeout and the value for the latency. If the value is undefined, we return undefined as it means the mirror is not yet measured.
+    normalizeLatencyValue(value) {
```
Hahihula: Since Rust sends Option<u32>, Tauri serialises None as null and Some(x) as x. So just use:

number → measured latency
null → timeout
(keep undefined as a unmeasured state)
no need for normalizeLatencyValue at all

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -109,44 +135,92 @@ export default {
...
-      return false;
-    },
-    get_available_pypi_mirrors: async function () {
-      const pypi_mirrors = await invoke("get_pypi_mirror_list", {});
-      this.pypi_mirrors = pypi_mirrors.mirrors.map((mirror, index) => {
-        return {
-          value: mirror,
-          label: mirror,
+        this.sortMirrorsByPing(this.idf_mirrors);
+        this.loading_idfs = this.mirrorsStore.loading_idf_urls || this.mirrorsStore.loading_idf_latency;
```
Hahihula: the loading should be also reactive computed properties:
```
loading_idfs = mirrorsStore.loading_idf_urls || mirrorsStore.loading_idf_latency
``

File: src/store.js [hunk 47]
```diff
@@ -427,3 +427,155 @@ export const useWizardStore = defineStore("wizard", {
...
+        this.loading_pypi_urls = false;
+      }
+
+      // Kick off progressive per-type background latency calculations
+      this.computeLatencyInBackground();
+    },
+
+    computeLatencyInBackground() {
+      const now = Date.now();
+      // IDF
```
Hahihula: i would really like if you can avoid this repetitive code ... maybe you can use something like:
```
const MIRROR_TYPES = ['idf', 'tools', 'pypi'];
state: () => {
    const base = {};
    for (const type of MIRROR_TYPES) {
      base[`${type}_urls`] = [];
      base[`${type}_latency_map`] = {};
      base[`selected_${type}`] = '';
      base[`loading_${type}_urls`] = false;
      base[`loading_${type}_latency`] = false;
      base[`${type}_last_updated`] = 0;
    }
    return {
      ...base,
      latency_ttl_ms: 15 * 60 * 1000,
    };
  },
```

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -172,9 +246,27 @@ export default {
...
-    this.get_available_idf_mirrors();
-    this.get_available_tools_mirrors();
-    this.get_available_pypi_mirrors();
+    // Initial load from store
+    this.refreshListFromStore('idf');
+    this.refreshListFromStore('tools');
+    this.refreshListFromStore('pypi');
+
+    // Subscribe to store updates and refresh progressively per-type
+    const unsub = this.mirrorsStore.$subscribe(() => {
```
Hahihula: you should avoid the $subscribe...use reactive computed properties
you can also avoid so much code repetition 
Please read more about Vuejs reactivity https://vuejs.org/guide/essentials/reactivity-fundamentals.html

src-tauri/src/gui/commands/installation.rs
```diff
@@ -253,21 +253,32 @@ app_handle: &AppHandle,
...
     });
 
-    let default_mirror = rust_i18n::t!("gui.installation.default_mirror").to_string();
-    let mirror = settings.idf_mirror.as_deref().unwrap_or(&default_mirror);
+    let default_mirror_str = rust_i18n::t!("gui.installation.default_mirror").to_string();
+    let is_simple_installation = app_state::is_simple_installation(&app_handle);
+    let mirror = settings.idf_mirror.as_deref().unwrap_or(&default_mirror_str);
+    let mut mirror_to_use: String = mirror.to_string();
+
+    if is_simple_installation && mirror == &default_mirror_str {
```
Hahihula: Care to explain this to me?
you are reading value from the localisation file:
`let default_mirror_str = rust_i18n::t!("gui.installation.default_mirror").to_string();`
and comparing it to this value?
How is this exactly supposed to work and in which languages?

src-tauri/src/gui/commands/installation.rs
```diff
@@ -770,11 +780,18 @@ fn is_process_running(pid: u32) -> bool {
     }
 }
 
-#[cfg(not(target_os = "windows"))]
 #[tauri::command]
+pub async fn start_installation_gui_cmd(app_handle: AppHandle) -> Result<(), String> {
```
Hahihula: You are unnecessarily adding another level of indirection...
this function should never exist...
You are obviously trying to achieve the default value... this is usually done in Rust by implementing the default trait for the structure
Please refactor your code according to this

src-tauri/src/lib/utils.rs
alirana01: done

src-tauri/src/lib/utils.rs
alirana01: yeah but why go over the loop once again if the head failed in that attempt why not break it right there.

src-tauri/src/lib/utils.rs
alirana01: done

src-tauri/src/cli/prompts.rs
alirana01: done

src-tauri/src/gui/utils.rs
alirana01: removing this as we dont need it anymore as moving to MirrorEntry vector the original latency calculation returns it sorted

src/main.js
alirana01: done

src/store.js
alirana01: Done

src/store.js
alirana01: removed

src/store.js
alirana01: done

src/store.js
alirana01: removed here as not needed now

src/components/wizard_steps/MirrorSelect.vue
alirana01: fixed

src/store.js
alirana01: good idea fixed

src/components/wizard_steps/MirrorSelect.vue
alirana01: updated

src-tauri/src/gui/commands/installation.rs
alirana01: 🙈🔫

src-tauri/src/gui/commands/installation.rs
alirana01: Done

src-tauri/src/lib/utils.rs
```diff
@@ -750,6 +776,130 @@ fn is_retryable_error(error: &io::Error) -> bool {
...
+    for url in mirrors.iter() {
+        if !head_latency_failed {
+            match measure_url_score_head(url, timeout).await {
+                Ok(score) => {
+                    info!("Mirror score: {} -> {}", url, score);
+                    mirror_entries.push(MirrorEntry { url: url.to_string(), latency: Some(score) });
+                }
+                Err(e) => {
+                    warn!("{}", e.to_string());
+                    head_latency_failed = true;
```
Hahihula, line 872: the break should be here

src-tauri/src/cli/prompts.rs
```diff
@@ -180,48 +181,96 @@ pub fn check_and_install_python(
...
+            })
+            .collect::<Vec<String>>();
+        let selected = generic_select(wizard_key, &display)?;
+        let url = selected.split(" (").next().unwrap_or(&selected).to_string();
+        set_value(config, url);
+    } else if needs_value {
+        let entries = calculate_mirrors_latency(candidates).await;
+        if let Some(entry) = entries.first() {
+            if entry.latency.is_none() {
+                info!("Selected {log_prefix} mirror: {} (timeout)", entry.url);
```
Hahihula: You just selected for the user mirror which timeouted and threat it as normal situation, with logging info ?

src-tauri/src/gui/utils.rs
```diff
@@ -40,6 +51,61 @@ pub fn is_path_empty_or_nonexistent(path: &str, versions: &[String]) -> bool {
...
+            }
+            None => {
+                log::info!("No calculated mirror found for {}, using fallback: {}", settings_key, fallback);
+                return fallback;
+            }
+        }
+    }
+}
+
+pub async fn get_mirror_to_use(mirror_type: MirrorType, settings: &Settings, is_simple_installation: bool, app_handle: &AppHandle) -> String {
```
Hahihula: every other function accepting app_handle has it as firt argument, can you pleare acomodate your to that custom for sake of readibility?

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -166,15 +198,72 @@ export default {
...
+        this.maybeAutoSelectBest('pypi');
+      }
+    }
+  },
   mounted() {
-    this.get_available_idf_mirrors();
-    this.get_available_tools_mirrors();
-    this.get_available_pypi_mirrors();
+    // Ensure background bootstrap runs if user navigates directly here before app init
+    if (
```
Hahihula: this means that if you have mirror of one type but not the others becaus for example network disconected before you finished measuring, you will not refetch them.

src/store.js
Hahihula: Could you please do the same for the rest of the `useMirrorsStore` ? You have the same repetitive patern all over it. My previos comment is clear hint how to achieve much less repetitive code.

src/store.js
```diff
@@ -427,3 +427,108 @@ export const useWizardStore = defineStore("wizard", {
     },
   },
 });
+
+export const useMirrorsStore = defineStore("mirrors", {
```
Hahihula, line 431: try to deduplicate the code in this store

src-tauri/src/lib/utils.rs
alirana01: Done

src-tauri/src/cli/prompts.rs
alirana01: fixed returning error if the top entry is none as that means some connection issue as vectors are sorted by latency and None (timeout) should be at the end not first entry

src-tauri/src/gui/utils.rs
alirana01: Done

src/components/wizard_steps/MirrorSelect.vue
alirana01: done

src/store.js
alirana01: done

src-tauri/src/cli/prompts.rs
```diff
@@ -217,12 +217,15 @@ where
...
-            } else {
+            if entry.latency.is_some() {
+                // The first entry is best mirror to select
                 info!("Selected {log_prefix} mirror: {} ({:?} ms)", entry.url, entry.latency.unwrap());
+                set_value(config, entry.url.clone());
             }
-            set_value(config, entry.url.clone());
+        } else {
+            // If the first entry is timeout or None there are no good mirrors to select try logging a proper message and return an error
+            info!("No good {log_prefix} mirrors found, please check your internet connection and try again§");
```
Hahihula: typo?

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -78,8 +102,9 @@
 </template>
 
 <script>
-import { ref } from "vue";
+import { } from "vue";
```
Hahihula: ?? What should this be?

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -93,63 +118,70 @@ export default {
...
-          label: mirror,
-        }
-      });
-      this.selected_idf_mirror = idf_mirrors.selected;
-      this.loading_idfs = false;
-      return false;
+    buildMirrorList(type) {
+      const urls = type === 'idf' ? this.mirrorsStore.idf_urls : type === 'tools' ? this.mirrorsStore.tools_urls : this.mirrorsStore.pypi_urls;
+      const entries = type === 'idf' ? this.mirrorsStore.idf_entries : type === 'tools' ? this.mirrorsStore.tools_entries : this.mirrorsStore.pypi_entries;
+      if (Array.isArray(entries) && entries.length > 0) {
```
Hahihula: you can use `entries?.length > 0`

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -93,63 +118,70 @@ export default {
...
-      this.idf_mirrors = idf_mirrors.mirrors.map((mirror, index) => {
-        return {
-          value: mirror,
-          label: mirror,
-        }
-      });
-      this.selected_idf_mirror = idf_mirrors.selected;
-      this.loading_idfs = false;
-      return false;
+    buildMirrorList(type) {
```
Hahihula, line 135: this function can be simplified using try catch block

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -93,63 +118,70 @@ export default {
...
-      this.pypi_mirrors = pypi_mirrors.mirrors.map((mirror, index) => {
-        return {
-          value: mirror,
-          label: mirror,
-        }
-      });
-      this.selected_pypi_mirror = pypi_mirrors.selected;
-      this.loading_pypi = false;
-      return false;
+    getBestMirror(list) {
```
Hahihula: using `reduce` this function is oneliner

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -93,63 +118,70 @@ export default {
...
+      if (candidates.length === 0) return null;
+      let best = candidates[0];
+      for (let i = 1; i < candidates.length; i++) {
+        if (candidates[i].ping < best.ping) best = candidates[i];
+      }
+      return best;
+    },
+    maybeAutoSelectBest(type) {
+      if (!this.autoSelect[type]) return;
+      const list = type === 'idf' ? this.idf_mirrors : type === 'tools' ? this.tools_mirrors : this.pypi_mirrors;
```
Hahihula: this can be written more readably like: `const list   = this[`${type}_mirrors`];`

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -93,63 +118,70 @@ export default {
...
+      const best = this.getBestMirror(list);
+      if (!best) return;
+      const selectedKey = type === 'idf' ? 'selected_idf_mirror' : type === 'tools' ? 'selected_tools_mirror' : 'selected_pypi_mirror';
```
Hahihula: const storeKey = `selected_${type}_mirror`;

src/components/wizard_steps/MirrorSelect.vue
```diff
@@ -93,63 +118,70 @@ export default {
...
+      const currentEntry = list.find(m => m.value === current) || null;
+      const currentPing = currentEntry ? currentEntry.ping : null;
+      const currentScore = (currentPing !== null && currentPing > 0) ? currentPing : Number.POSITIVE_INFINITY;
+      if (best.ping < currentScore) {
+        this[selectedKey] = best.value;
+      }
     },
     isDefaultMirror(mirror, type) {
-      return mirror === this.defaultMirrors[type];
+      const selectedFromStore = type === 'idf' ? this.mirrorsStore.selected_idf : type === 'tools' ? this.mirrorsStore.selected_tools : this.mirrorsStore.selected_pypi;
```
Hahihula: similar as abow this can be more readable.

src/store.js
```diff
@@ -427,3 +427,134 @@ export const useWizardStore = defineStore("wizard", {
...
+  state: () => ({
```
Hahihula, line 432: Unified the structure for all mirror types could make the store much more readable and maintainable.
for example like this:
```
mirrors: {
      idf: {
        urls: [],
        entries: [],
        selected: "",
        loading_urls: false,
        loading_latency: false,
        last_updated: 0,
      },
      tools: {
        urls: [],
        entries: [],
        selected: "",
        loading_urls: false,
        loading_latency: false,
        last_updated: 0,
      },
      pypi: {
        urls: [],
        entries: [],
        selected: "",
        loading_urls: false,
        loading_latency: false,
        last_updated: 0,
      },
    },

```
it will also simplyfy the code in the rest of the store and the component...

src/components/wizard_steps/MirrorSelect.vue
alirana01: Simplified with new simplified code

src-tauri/src/cli/prompts.rs
alirana01: fixed

src/components/wizard_steps/MirrorSelect.vue
alirana01: originally i see ref import from master but not sure if its even used there but i reverted back to that to avoid any issues

src/components/wizard_steps/MirrorSelect.vue
alirana01: simplified not exactly try catch but kindly take a look if it is okay now

src/components/wizard_steps/MirrorSelect.vue
alirana01: done

src/components/wizard_steps/MirrorSelect.vue
alirana01: done

src/components/wizard_steps/MirrorSelect.vue
alirana01: done

src/components/wizard_steps/MirrorSelect.vue
alirana01: done

src/store.js
alirana01: Done

src/components/wizard_steps/MirrorSelect.vue
Hahihula: Thats not what i ment. It is ok to remove unused import, but why did you kept empty brackets there?

Fabricio-ESP (approved):
Confirmed working.
Thanks for implementing this feature.

Hahihula (approved):
LGTM. Please do not forget to squash the commits.

PR #289, opened 2025-12-01
Eim 377 fix new version check

Hahihula:
This pr moves checking of the new release to the rust part from the js part
it also unifies calling place for prerequisities check and platform information


Fabricio-ESP (approved):
Looks good.

PR #293, opened 2025-12-02
update eim select cli command to output activation instruction

Hahihula:


Fabricio-ESP (approved):
Nice improvement, thanks

PR #297, opened 2025-12-04
Update piepeline with winget repo sync

Hahihula:
the winget fork repository needs to besynce before the PR is created

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -176,6 +176,7 @@ jobs:
           upload_url: ${{ github.event.release.upload_url }}
           asset_path: release_cli/${{ matrix.package_name }}/eim.zip
           asset_name: eim-cli-${{ matrix.package_name }}.zip
+
       # offline installer
       - name: Build offline_installer_builder
         run: |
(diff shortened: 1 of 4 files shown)


src/components/UpdateNotification.vue
```diff
@@ -188,8 +188,10 @@ export default {
         return 'sudo apt update && sudo apt upgrade eim'
       } else if (os.includes('darwin') || os.includes('macos')) {
         return 'brew update && brew upgrade eim'
+      } else if (os.includes('windows')) {
+        return 'winget upgrade Espressif.eim'
       }
-      return null // Windows doesn't have command-line update
+      return null
     })
```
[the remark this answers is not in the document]
alirana01: @Hahihula looking at this maybe we can do something like this to have only one function for computation given that they are both interlinked 
```javascript
const updateInfo = computed(() => {
  const os = currentPlatform.value.toLowerCase();

  if (os.includes('linux')) {
    return {
      command: 'sudo apt update && sudo apt upgrade eim',
      description: t('app.updateAvailable.updateWithApt'),
    };
  }

  if (os.includes('darwin') || os.includes('macos')) {
    return {
      command: 'brew update && brew upgrade eim',
      description: t('app.updateAvailable.updateWithBrew'),
    };
  }
...............
```

src/components/UpdateNotification.vue
```diff
@@ -199,6 +201,8 @@ export default {
         return t('app.updateAvailable.updateWithApt')
       } else if (os.includes('darwin') || os.includes('macos')) {
         return t('app.updateAvailable.updateWithBrew')
+      } else if (os.includes('windows')) {
+        return t('app.updateAvailable.updateWithWinget')
```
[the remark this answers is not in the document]
alirana01: I think there is missing field in locales probably or not sure how they are calculated

Fabricio-ESP (approved):
Looks good.

.github/workflows/build.yaml
```diff
@@ -259,11 +259,18 @@ jobs:
 
       - name: Upload CLI .rpm artifact
         if: github.event_name == 'release' && github.event.action == 'created'
```
Fabricio-ESP: We could remove this, so we generate the rpm package on every run, and leave the filter for the " Upload CLI .rpm Release Asset" step only. Like we do for the GUI file.

.github/workflows/build.yaml
Fabricio-ESP: Same here, we can remove this conditional and have the cli deb file build on every run, if we want to test it.

Fabricio-ESP (dismissed):

Fabricio-ESP (dismissed):
Looks good.

.github/workflows/build.yaml
Hahihula: good point, thanks for the catch

.github/workflows/build.yaml
Hahihula: ^^ same here... thanks...

PR #305, opened 2025-12-12
EIM-123 Added minimal tools selection implelmentation

Hahihula:
2025-12-11717136157 082498-0100 - DEBUG - starting new comectles Mtpsiraw githubusercentent com

image

### DIFF
--- a/.github/workflows/update-homebrew.yml
+++ b/.github/workflows/update-homebrew.yml
@@ -119,13 +119,6 @@ jobs:
               end
             end
 
-            # Runtime dependencies for QEMU (used by ESP-IDF for emulation)
-            depends_on "libgcrypt"
-            depends_on "glib"
-            depends_on "pixman"
-            depends_on "sdl2"
-            depends_on "libslirp"
-
             # DFU utility for flashing
             depends_on "dfu-util"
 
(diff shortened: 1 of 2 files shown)


alirana01 (dismissed):

Fabricio-ESP:
After looking into this PR, I have a couple of notes observed on a MAC arm machine:

On the GUI screen before the tools selection, the button description needs to be updated, since we have added a new step. Also note the missing translation tag:
Untitled
On the tools selection screen, I noticed that clocking right on the checkbox does not activate the event listener, just when clicking on the box, but outside the checkbox.
I noticed that on GUI all tools are selected by default, while the CLI they are unselected by default. Can we set both the same? Maybe all optional tools unselected by default?

I proceeded with the installation and de-selected QEMU, then I ended up with a pre-requisites error for QEMU (but it was de-selected)
Untitled 2
On the logs I have just the list of pre-requisites indicating they are missing (and indeed they are not installed).
Installing the pre-requisites before running the install resulted on QEMU being actually installed, even if it was not selected on the GUI.

On CLI the tools selection worked as expected.

Similar behaviour on Linux.

Windows worked as expected.

Hahihula:
the buttons should be fixed now:
image

Fabricio-ESP (approved):
Installation now works as expected for CLI and GUI, QEMU is not selected on default installation but available to users.

PR #310, opened 2025-12-16
reset settings after finished installation

Hahihula:
now the setting chosen by user during the expert installation should not persist to other installations


Fabricio-ESP (approved):
Confirmed Fix.

PR #319, opened 2025-12-29
feat: Add codepell check

Shen7436:
This PR adds a `codespell` job to the workflow checks, preventing possible typos from being committed. 
The following file types and paths are excluded from the check:

-  `*.lock`
- `*.rs`
- `*.vue`
- `*.js`
- `*.json`
- `*.ps1`
- files in `build` and `man` folders
- src-tauri/locales/app.yml

### DIFF
--- a/.github/workflows/pre_check.yml
+++ b/.github/workflows/pre_check.yml
@@ -1,0 +1,47 @@
+name: Spell Check (typos)
+
+# Triggers on pull requests and pushes to master or main branches
+on:
+  pull_request:
+  push:
+    branches: [master, main]
+
+jobs:
+  typos:
+    runs-on: ubuntu-latest
+    steps:
+      # Fetches full Git history to enable branch comparison
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Identifies changed files using git diff
+      - name: Get modified files
+        id: modified_files
+        run: |
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            BASE_BRANCH="${{ github.base_ref }}"
+            git fetch origin $BASE_BRANCH --depth=1
+            MODIFIED_FILES="$(git diff --name-only origin/$BASE_BRANCH...HEAD || echo "")"
+          else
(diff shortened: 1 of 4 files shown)


src-tauri/locales/app.yml
```diff
@@ -59,7 +59,7 @@ wizard.instalation_path.unselected:
 wizard.idf.mirror:
   en: Select the source from which to download ESP-IDF
   cn: 请选择下载 ESP-IDF 的来源
-wizard.idf.sucess:
```
Hahihula, line 62: the former key is still used in the code. Change cannot be made just in the language file, the app will ju st stop working or in the better case print out the key instead of the message.

Hahihula (commented):
both codespell and precommit are purely python things. This is not a python repository. You are introducing dependency on completly newe language. Developers can not be expected to have working python environment. I thing we need to discuss if this is worthy or if there is othr solution.
Also, changing keys in language file (which should never be displayed to the user so are not userfacing strings) and not changing the code invoking them means the app could brake or in the best case display the key insted of the message. This is really not correct way to do this.
based on the files excluded looks like you are interested in the spellchecking of the documentation. but even if we want to include spellchecking of the codefiles, probalby this CI action would be more usefull
https://github.com/marketplace/actions/typos-action

src-tauri/locales/app.yml
Shen7436: Thanks for the reminder! 

I attempted to use regex patterns in `.typos.toml` to selectively ignore only the YAML keys while still checking the translation content. Unfortunately, `typos-cli` doesn't support this – the `extend-ignore-identifiers-re` option only works for code identifiers, not YAML keys. As a result, I've excluded the entire `app.yml` file from typo checking.

Shen7436:
Hi @Hahihula Thanks for your suggestions! I've switched to using `typos-action` instead of `codespell` to avoid the Python dependency.

I've also properly configured `.typos.toml` based on the former `.codespellrc` and added `app.yml` to the ignore list.

The CI now uses typos-action to check documentation while excluding some code files. Could you please take another look when you have a chance? Thx:)

Hahihula (approved):
Thanks for the changes.

PR #325, opened 2026-01-05
EIM-397 easy installation now by default includes the IDE feature

Hahihula:


src-tauri/src/gui/commands/installation.rs
```diff
@@ -1022,6 +1022,41 @@ pub async fn start_simple_setup(app_handle: tauri::AppHandle) -> Result<(), Stri
...
     }
 
+    match update_settings(&app_handle, |settings| {
+      let mut features = settings.idf_features
+        .clone()
+        .unwrap_or_default();
+
+      if !features.contains(&"ide".to_string()) {
+        features.push("ide".to_string());
+      }
```
alirana01, line 1032: I am a bit confused can't we have this as a default value with the features vector? 🤔

alirana01 (approved):
I just have a small question about this if its fine as it is it can be merged

src-tauri/src/gui/commands/installation.rs
Hahihula: we actually do not want it to be default for every occasion, only for the easy install where user do not have the possibility to select the features.

PR #326, opened 2026-01-05
EIM-357 fixed the crash when invalid config is provided. 

Hahihula:
Now simple message is displayed to the user

### DIFF
--- a/src/components/BasicInstaller.vue
+++ b/src/components/BasicInstaller.vue
@@ -273,12 +273,16 @@ export default {
         }
 
         if (selected) {
-          const _ = await invoke("load_settings", { path: selected });
-          message.success('Configuration loaded')
-          router.push('/installation-progress')
+          invoke("load_settings", { path: selected }).then(() => {
+            message.success(t('basicInstaller.messages.configLoaded'))
+            router.push('/installation-progress')
+          }).catch((error) => {
+            console.error('Failed to load configuration:', error)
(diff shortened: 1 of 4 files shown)


src/components/BasicInstaller.vue
```diff
@@ -273,9 +273,13 @@ export default {
         }
 
         if (selected) {
-          const _ = await invoke("load_settings", { path: selected });
-          message.success('Configuration loaded')
-          router.push('/installation-progress')
+          invoke("load_settings", { path: selected }).then(() => {
+            message.success('Configuration loaded')
```
alirana01: if this is not a console log shouldn't we externalize this as well

src/components/BasicInstaller.vue
```diff
@@ -273,9 +273,13 @@ export default {
...
+            router.push('/installation-progress')
+          }).catch((error) => {
+            console.error('Failed to load configuration:', error)
+            message.error('Failed to load configuration')
```
alirana01: Same here

src/components/BasicInstaller.vue
Hahihula: oh... good catch, thanks

src/components/BasicInstaller.vue
Hahihula: Thanks. Added to the i18n file

PR #335, opened 2026-01-13
improving gui error message so it's more clear for the users

Hahihula:
The message as displayed in the GUI was not clear enough. Hopefully this message is more informative.


alirana01 (approved):
LGTM! Small question
Shouldn't we mention the version to the user to install or a range as per gemini's review?
Is this specifically by design to not share the version?

PR #342, opened 2026-01-15
EIM-334: prerequisities check enhancement

alirana01:
## Description

This PR enhances the prerequisites checking system to handle scenarios where the system cannot execute shell commands or encounters errors during verification. Previously, if the shell failed or a system error occurred, users were stuck with no way to proceed.
image

image

### Backend (Rust)

- Added `verify_shell_execution()` to test basic shell execution on each platform:
  - **Linux:** `sh -c "echo test"`
  - **macOS:** `zsh -c "echo test"`
  - **Windows:** `cmd /c echo test`
- Added `PrerequisitesCheckResult` struct with:
  - `missing: Vec<&'static str>` — list of missing prerequisites
  - `can_verify: bool` — whether the system can run verification commands
  - `shell_failed: bool` — whether shell execution specifically failed
- Added `check_prerequisites_with_result()` to return detailed verification status
- Updated CLI to show an appropriate error message mentioning the `--skip-prerequisites-check` flag
- Updated GUI backend to return detailed JSON with `can_verify` and `shell_failed` fields

### Frontend (Vue)

- Updated `SimpleSetup.vue` (Easy Installation) to show a skip button on verification failure
- Updated `PrerequisitesCheck.vue` (Expert Installation) to stop auto-navigation and show a skip option
- Fixed wizard auto-navigation bug that was skipping when shell failed (empty `missing` array)

### Localization

- Added GUI-specific locale keys in `app.yml` for shell verification failures
- Consolidated duplicate prerequisite messages into `common.prerequisites` in `en.json` and `cn.json`

## Checklist

### DIFF
--- a/src-tauri/src/cli/prompts.rs
+++ b/src-tauri/src/cli/prompts.rs
@@ -93,7 +93,8 @@ pub fn check_and_install_prerequisites(
             // Some prerequisites are missing
             let unsatisfied_prerequisites: Vec<String> = 
                 result.missing.into_iter().map(|p| p.to_string()).collect();
-            
+
+            info!("{} {:?}", t!("prerequisites.missing"), unsatisfied_prerequisites);
             info!(
                 "{}",
                 t!(


alirana01 (commented):
Initial self review

src-tauri/src/lib/system_dependencies.rs
```diff
@@ -438,13 +515,53 @@ fn check_tools_installed(tools: Vec<&'static str>) -> Result<Vec<&'static str>,
 ///
 /// * `Ok(Vec<&'static str>)` - If the function completes successfully, returns a vector of unsatisfied tools.
 /// * `Err(String)` - If an error occurs, returns an error message.
-pub fn check_prerequisites() -> Result<Vec<&'static str>, String> {
+// pub fn check_prerequisites() -> Result<Vec<&'static str>, String> {
```
Hahihula: what is this commented out code?

src-tauri/src/gui/commands/prequisites.rs
```diff
@@ -19,72 +19,92 @@ pub fn get_prequisites() -> Vec<&'static str> {
 /// Checks which prerequisites are missing
 #[tauri::command]
 pub fn check_prequisites(app_handle: AppHandle) -> Vec<String> {
```
alirana01, line 20: @Hahihula 
The check_prequisites() function returns an empty Vec<String> for both "all prerequisites satisfied" and "verification failed" cases.
In installation.rs (lines 873, 887, 898, 1543), this means if shell verification fails, the code proceeds as if prerequisites are satisfied (a warning is shown to the user, but installation continues).
Is this silent pass-through acceptable for the installation flow? Or should we update these callers to use check_prerequisites_detailed() which distinguishes between "success" and "can't verify"?

Hahihula: In case it's really working as you described than it's definitely bug, but this task was filed because when the check failed there was no way to proceed so i am not sure it's actually working the way you are describing. I will take a closer look.
But you are completely right. we should not silent pass.

src-tauri/src/lib/system_dependencies.rs
```diff
@@ -7,6 +7,35 @@ use crate::{command_executor, utils::find_by_name_and_extension};
 
 pub const PYTHON_NAME_TO_INSTALL: &str = "python313";
 
+/// Result of a prerequisites check operation.
+///
+/// This struct provides detailed information about the outcome of checking
+/// system prerequisites, including both the list of missing tools and whether
+/// the system was able to run verification commands at all.
+#[derive(Debug, Clone)]
+pub struct PrerequisitesCheckResult {
```
Hahihula, line 16: aren't the can_verify and shell_failed mutualy exclusive ? maybe we do not need the shell_failed because that is what can_verify=false means ?

alirana01: `shell_failed` distinguishes **why** `can_verify` is false:

- `can_verify=false, shell_failed=true` → Shell commands don't work on this system
- `can_verify=false, shell_failed=false` → Shell works, but another error occurred (e.g., unsupported package manager)

This allows the UI to show appropriate messages:
- Shell failed → "Basic shell execution failed"
- Other error → "A system error occurred: {specific error}"

Without `shell_failed`, we couldn't differentiate between these two failure modes.
So basically to relay proper information to user

src-tauri/src/lib/system_dependencies.rs
alirana01: Removed

src-tauri/src/lib/system_dependencies.rs
Hahihula: but than your logic is flawed ... you return 
verify_shell_execution -> false for both unsupported OS as well as for failed shell execution
in current implementation they are mutualy exclusive

src/components/SimpleSetup.vue
```diff
@@ -230,6 +233,44 @@ export default {
...
+
+    // Helper function to fetch settings, versions, and validate installation path
+    const prepareForInstallation = async () => {
+      try {
+        const settings = await invoke('get_settings')
+        installPath.value = settings?.path
+        console.log('Default installation path:', installPath.value)
+
+        const versions = await invoke('get_idf_versions', { includeUnstable: false })
+        selectedVersion.value = versions?.[0]?.name || 'v5.5.1'
```
Hahihula: i know it was here already, but as you are refactoring this function toss this hardcoded default out. if we could not get_idf_versions, it should fail.

Hahihula (changes requested):
under current logic the can_verify and shell_failed are mutually exclusive, either your logic is flawed or their are redundant. 
Also could you please finally stick to the branch naming convention in this repository in next PRs ?

src/components/SimpleSetup.vue
alirana01: Done

src-tauri/src/lib/system_dependencies.rs
alirana01: Yep fixed now need your expert review again please. Callers are already handling this properly

Hahihula (dismissed):
Thanks!

PR #348, opened 2026-01-19
EIM-344 component manager integration to offline installer

Hahihula:
Both online and offline installers should now sync components from the component manager. 
as a fallback for older versions and because if the sync fails, the components will be downloaded as part of first build, the failure is graceful with only a warning.

### DIFF
--- a/src-tauri/src/cli/wizard.rs
+++ b/src-tauri/src/cli/wizard.rs
@@ -390,7 +390,15 @@ pub async fn run_wizzard_run(mut config: Settings) -> Result<(), String> {
         // copy IDFs
         copy_idf_from_offline_archive(archive_dir, &config)?;
         let compoments_dir = PathBuf::from(config.tool_install_folder_name.clone().expect("Tools install folder not defined"));
-        copy_components_from_offline_archive(archive_dir, &compoments_dir)?;
+        match copy_components_from_offline_archive(archive_dir, &compoments_dir) {
+            Ok(_) => {
+                info!("{}", t!("wizard.components_copy.success"));
+            }
+            Err(err) => {
+                warn!("{}: {}", t!("wizard.components_copy.failure"), err);
+                // The component will be donwloaded during first build and not all of them are actually needed but most importantly this is for backward compatibility with old IDF not supporting the new component manager, so we don't want to block the installation if the copy fails
+            }
+        }
     }
 
     // select target & idf version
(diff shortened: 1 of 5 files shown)


src-tauri/src/lib/mod.rs
```diff
@@ -1678,6 +1680,26 @@ pub fn single_version_post_install(
...
+    // Synchronize component manager
+    info!(
+        "Syncing components to {:?}...",
+        tool_install_directory
+    );
+    let command = format!(
+        "compote registry sync --resolution=latest --recursive {}",
+        tool_install_directory
+    );
+    match version_manager::run_command_in_context(&idf_path, &command) {
```
alirana01: **Question about `run_command_in_context` timing:**

I was looking at how `run_command_in_context` works in `version_manager.rs` it calls `list_installed_versions()` and tries to match the identifier against already-registered installations. Since `single_version_post_install` is called *during* installation (before the activation script and version config are written, which happens further down in this same function), I'm wondering whether `run_command_in_context` would be able to find the version at this point.

If the version isn't registered yet it would hit the `Err("Version {} not installed")` path, and the sync would silently fall through to the `warn!` on line 1701. The install would still succeed but the component sync wouldnot actually run not because it couldn't but becasue it cant find the version.

Could you check if this is the case? If so, one option might be to use a direct execution approach (like `execute_command_with_dir` with the right env vars) similar to how the offline builder handles it — that way it doesn't depend on the version being registered first.

Happy to be wrong on this just wanted to flag it in case.

src-tauri/src/cli/wizard.rs
```diff
@@ -388,6 +389,8 @@ pub async fn run_wizzard_run(mut config: Settings) -> Result<(), String> {
         let archive_dir = offline_archive_dir.as_ref().unwrap();
         // copy IDFs
         copy_idf_from_offline_archive(archive_dir, &config)?;
+        let compoments_dir = PathBuf::from(config.tool_install_folder_name.clone().expect("Tools install folder not defined"));
+        copy_components_from_offline_archive(archive_dir, &compoments_dir)?;
```
alirana01: We must be catching this error instead of propagating it, thinking if someone tries to use an older archive with this it will fail? 🤔

src-tauri/src/lib/mod.rs
Hahihula: Thanks for the catch! 
You are absolutely right, we need to first finish writing the eim_idf.json file.
I missed it as i have second bug in the tests and write multiple times the same path. 
Good eye.

src-tauri/src/cli/wizard.rs
Hahihula: If someone tries to use an older offline archive, it should fail. That's why we are purging old archives with every release.
Also, the sad part is that as the components are migrated to the registry without this step the installed IDF would be useless.

src-tauri/src/lib/mod.rs
Hahihula: Should be fixed now.

XDanielPaul:
Tested, LGTM!

Issue #366, opened 2026-01-25
EIM crash during the "Python Sanity Check" step (EIM-448)


hsmptg:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

not having a crash

### Actual behavior (suspected bug)

PS C:\Users\Helio> & 'C:\Program Files\eim\eim.exe' gui
Running GUI...
PS C:\Users\Helio> [2026-01-25][18:04:56][eim::gui::commands::utils_commands][INFO] Set locale to: en
[2026-01-25][18:04:56][eim::gui::commands::utils_commands][INFO] Set locale to: en

thread 'main' (3024) panicked at src\lib\python_utils.rs:823:67:
called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 8, error_len: Some(1) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' (3024) panicked at library\core\src\panicking.rs:230:5:
panic in a function that cannot unwind

### Error logs or terminal output

```shell

```

### Steps to reproduce the behavior

1. run EIM (GUI or CLI)
2. during the "Python Sanity Check" step it Crashes!

### Project release version

v0.6.0 and v0.7.0

### System architecture

Intel 64-bit (modern PC)

### Operating system

Windows

### Operating system version

Windows 11

### Shell

PowerShell

PR #370, opened 2026-01-27
EIM-443 handle space in python path

Hahihula:
now space in python or scoop path should no longer be problem.

### DIFF
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -77,6 +77,8 @@ pub fn setup_cli(
     // Module filters are applied globally
     Dispatch::new()
         .format(formatter)
+        // Filter reqwest to only show warnings and errors
+        .level_for("reqwest", LevelFilter::Warn)
         // Apply file at Trace level
         .chain(
             Dispatch::new()
--- a/src-tauri/src/lib/command_executor.rs
+++ b/src-tauri/src/lib/command_executor.rs
@@ -71,43 +71,32 @@ impl CommandExecutor for DefaultExecutor {
             .spawn()
     }
     fn run_script_from_string(&self, script: &str) -> std::io::Result<Output> {
-        self.execute("bash", &["-c", script])
+        Command::new("bash")
+            .args(["-c", script])
+            .output()
     }
 }
 
 #[cfg(target_os = "windows")]
 struct WindowsExecutor;
 
-/// Retrieves the major version number of PowerShell installed on the system.
-///
-/// This function executes a PowerShell command to fetch the major version number
-/// of the installed PowerShell. On Windows, it uses the CREATE_NO_WINDOW flag
-/// to prevent a console window from appearing during execution.
-///
-/// # Returns
-///
-/// Returns a `Result` containing:
-/// - `Ok(i32)`: The major version number of PowerShell if successfully retrieved.
-///   If parsing fails, it defaults to version 5.
-/// - `Err(std::io::Error)`: An error if the PowerShell command execution fails.
-///
-/// # Platform-specific behavior
-///
-/// On Windows, this function uses the CREATE_NO_WINDOW flag to suppress the console window.
 pub fn get_powershell_version() -> std::io::Result<i32> {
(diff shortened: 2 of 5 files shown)


alirana01 (commented):
## Question about the dual execution approach

Thanks for the fix! I have a question about the design:

With the new escaping in `execute_command()`, shell features like variable expansion (`$HOME`) and globs are now blocked. Given that:

| Function | Shell overhead | Handles spaces | Shell features |
|----------|---------------|----------------|----------------|
| `execute_command()` | Yes | Yes (via escaping) | No (blocked by escaping) |
| `execute_command_direct()` | No | Yes (native) | No |

**Question:** What's the benefit of keeping `execute_command()` going through bash/PowerShell if escaping prevents shell features anyway? It seems like we now have shell overhead without shell benefits.

### Suggestions

1. Could `execute_command()` just call `execute_command_direct()` internally? The only difference now is PATH resolution and shell environment.

2. Only `create_python_venv()` was updated to use `execute_command_direct()`. Other functions in `python_utils.rs` that take python paths could also have space issues:
   - `detect_python_version()` (line 401)
   - `run_python_script()` (line 786)
   - `get_python_version()` (line 818)

My apologies for asking clarifications in review

Hahihula:
> ## Question about the dual execution approach
> Thanks for the fix! I have a question about the design:
> 
> With the new escaping in `execute_command()`, shell features like variable expansion (`$HOME`) and globs are now blocked. Given that:
> 
> Function	Shell overhead	Handles spaces	Shell features
> `execute_command()`	Yes	Yes (via escaping)	No (blocked by escaping)
> `execute_command_direct()`	No	Yes (native)	No
> **Question:** What's the benefit of keeping `execute_command()` going through bash/PowerShell if escaping prevents shell features anyway? It seems like we now have shell overhead without shell benefits.
> 
> ### Suggestions
> 1. Could `execute_command()` just call `execute_command_direct()` internally? The only difference now is PATH resolution and shell environment.
> 2. Only `create_python_venv()` was updated to use `execute_command_direct()`. Other functions in `python_utils.rs` that take python paths could also have space issues:
>    
>    * `detect_python_version()` (line 401)
>    * `run_python_script()` (line 786)
>    * `get_python_version()` (line 818)
> 
> My apologies for asking clarifications in review

when you are saying `Shell features` you meant expansion of $HOME variable, but the diferences are with lot more variables and environment settings. Generaly we should be migrating to the execute_command_direct but it is out of scope of these change to do this refactoring for whole codebase.

alirana01:
> > ## Question about the dual execution approach
> > Thanks for the fix! I have a question about the design:
> > With the new escaping in `execute_command()`, shell features like variable expansion (`$HOME`) and globs are now blocked. Given that:
> > Function	Shell overhead	Handles spaces	Shell features
> > `execute_command()`	Yes	Yes (via escaping)	No (blocked by escaping)
> > `execute_command_direct()`	No	Yes (native)	No
> > **Question:** What's the benefit of keeping `execute_command()` going through bash/PowerShell if escaping prevents shell features anyway? It seems like we now have shell overhead without shell benefits.
> > ### Suggestions
> > 
> > 1. Could `execute_command()` just call `execute_command_direct()` internally? The only difference now is PATH resolution and shell environment.
> > 2. Only `create_python_venv()` was updated to use `execute_command_direct()`. Other functions in `python_utils.rs` that take python paths could also have space issues:
> >    
> >    * `detect_python_version()` (line 401)
> >    * `run_python_script()` (line 786)
> >    * `get_python_version()` (line 818)
> > 
> > My apologies for asking clarifications in review
> 
> when you are saying `Shell features` you meant expansion of $HOME variable, but the diferences are with lot more variables and environment settings. Generaly we should be migrating to the execute_command_direct but it is out of scope of these change to do this refactoring for whole codebase.

and do we need to update the calls in these 

> >    * `detect_python_version()` (line 401)
> >    * `run_python_script()` (line 786)
> >    * `get_python_version()` (line 818)

as I believe one of the recent issue #366 could also be related to this

PR #376, opened 2026-01-28
EIM-456 run command in context

Hahihula:
We are adding command to run some command in IDF shell context
more info in the included documentation update


kolipakakondal (approved):
LGTM.

PR #402, opened 2026-02-06
EIM-89: eim select added for the activation script

alirana01:
## Description

Added the eim select in sourcing scripts. This should select the correct version for eim and not have issues as mentioned in Jira ticket [EIM-89](https://jira.espressif.com:8443/browse/EIM-89)

**This is supposed to work if the eim is in path and fail silently as well.**

## Testing

Run the activation script and verify that eim select runs
also make sure that eim is in path.

We need to verify this with eim not in Path to see that the sourcing works for activation script and no errors are shown we have to fail silently in case of errors for eim select

---

## Checklist

### DIFF
--- a/src-tauri/powershell_scripts/idf_tools_profile_template.ps1
+++ b/src-tauri/powershell_scripts/idf_tools_profile_template.ps1
@@ -3,6 +3,9 @@ param(
     [switch]$e
 )
 
+# Capture absolute eim path early (before PATH changes)
+$_EimBinPath = (Get-Command eim -ErrorAction SilentlyContinue).Source
+
 {{env_var_pairs}}
 
 function Parse-CMakeVersion {
@@ -130,3 +133,11 @@ Write-Host 'parttool.py'
 Write-Host ''
 Write-Host 'Python environment activated.'
 Write-Host 'You can now use IDF commands and Python tools.'
+
+# Sync selection with eim_idf.json for IDEs (silent on failure)
+if ($_EimBinPath) {
+    try {
+        & $_EimBinPath select "{{idf_version}}" 2>$null
+        if ($LASTEXITCODE -eq 0) { Write-Host "eim select {{idf_version}}" }
+    } catch {}
+}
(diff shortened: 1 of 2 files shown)


src-tauri/powershell_scripts/idf_tools_profile_template.ps1
```diff
@@ -130,3 +133,11 @@ Write-Host 'parttool.py'
 Write-Host ''
 Write-Host 'Python environment activated.'
 Write-Host 'You can now use IDF commands and Python tools.'
+
+# Sync selection with eim_idf.json for IDEs (silent on failure)
+if ($_EimBinPath) {
+    try {
+        & $_EimBinPath select "{{idf_version}}" 2>$null
```
Hahihula, line 140: I am almost sure the syntax `2>$null` was introduced in PS5 (or was buggy before and was fixed in v5) and may not work on PS4. Could you please verify?
But as it's inside a try-catch block, this should not be a real issue, just a nitpick.

src-tauri/powershell_scripts/idf_tools_profile_template.ps1
```diff
@@ -3,6 +3,9 @@ param(
     [switch]$e
 )
 
+# Capture absolute eim path early (before PATH changes)
+$_EimBinPath = (Get-Command eim -ErrorAction SilentlyContinue).Source
```
Hahihula, line 7: If no EIM is present, you are effectively calling `null.Source`, which will fail and probably not silently. Can you check this?

alirana01: This is working fine I have verified it and -ErrorAction flag is supposed to skip this as it will simply return the PS variable $null.

src-tauri/powershell_scripts/idf_tools_profile_template.ps1
alirana01: $null is part of Powershell since v1.0 and since v3.0 you can basically redirect error stream using this

PR #415, opened 2026-02-09
EIM-452: debug archives purging workflow

alirana01:
## Description

Implements the offline installer archive purging strategy [EIM-452](https://jira.espressif.com:8443/browse/EIM-452): purge unsupported/EOL archives on release and add a scheduled job to clean the debug directory on S3.

### 1. Version-aware purge on release (`build_offline_installer_archives.yaml`)

- **Before:** "Purge old archives" deleted any `archive_*.zst` under `dl/eim/` older than 24 hours. Unsupported (e.g. EOL) versions could remain if recently touched.
- **After:** "Purge unsupported archives from S3" runs when `should_purge` is true (release or manual `purge_all`):
  - Builds the allowed set from **build-info artifacts** (exact filenames of archives built in this run).
  - Lists all `archive_*.zst` under `dl/eim/archive_` (excludes `debug/`).
  - Deletes any archive not in the allowed set and its corresponding `.log` file.
  - Appends deleted filenames to `old_files.txt` so "Merge and update JSON" removes them from `offline_archives.json`.
- "Download all build infos" was moved **before** the purge step so the allowed set is available.
- No change to when the workflow runs: still only invoked from `build.yaml` on **release** (and optionally via `workflow_dispatch` with `purge_all`).

### 2. Scheduled purge of debug directory (`purge_debug_offline_archives.yml`)

- **New workflow** to prevent unbounded growth of `s3://espdldata/dl/eim/debug/`.
- **Triggers:** `schedule` (daily at 04:00 UTC) and `workflow_dispatch`.
- **Behaviour:** Lists objects under `dl/eim/debug/`, deletes those older than 48 hours. No changes to `offline_archives.json`.
- **Inputs (manual run):** `dry_run` (default `false`), `max_age_hours` (default `48`).
- Uses the same AWS credentials pattern as other S3 workflows.

### 3. Temporary test workflow (`test_purge_unsupported_archives.yml`)

- **Purpose:** Test the version-aware purge logic on a branch without deleting anything.
- **Triggers:** `push` to `offline_purging_strategy` and `workflow_dispatch`.
- Fetches supported versions from `idf_versions.json` (or uses `versions_override`), builds the allowed set with the same naming as production (`archive_v${version}_${platform}.zst`), lists S3, and reports what **would** be deleted/kept. Read-only; no `aws s3 rm`.
- **To do before merge:** Remove this file.

## Testing

- Dry-run test workflow was run on the branch; with the corrected filename construction (`archive_v${version}_...` so S3 names like `archive_vv5.1.7_...` match), it correctly classifies archives as "would delete" vs "would keep".
- Real purge runs only on release (or manual run with `purge_all`).

### DIFF
--- a/.github/workflows/build_offline_installer_archives.yaml
+++ b/.github/workflows/build_offline_installer_archives.yaml
@@ -444,30 +444,34 @@ jobs:
           echo ""
 
           # List all .zst objects under dl/eim/archive_ (excludes debug/ since those are dl/eim/debug/archive_)
+          # With --output text, empty string is returned when no objects (not "None")
           S3_OBJECTS=$(aws s3api list-objects-v2 \
             --bucket "${BUCKET}" \
             --prefix "${PREFIX}" \
             --query 'Contents[?ends_with(Key, `.zst`)].[Key,Size]' \
-            --output text 2>/dev/null || true)
+            --output text 2>&1) || { echo "::warning::Failed to list S3 objects. Check bucket=${BUCKET} prefix=${PREFIX} and AWS credentials."; S3_OBJECTS=""; touch old_files.txt; }
 
           touch old_files.txt
           DELETED_COUNT=0
           KEPT_COUNT=0
 
-          if [ -n "${S3_OBJECTS}" ] && [ "${S3_OBJECTS}" != "None" ]; then
+          if [ -n "${S3_OBJECTS}" ]; then
             while IFS=$'\t' read -r key size; do
               [ -z "${key}" ] && continue
               filename=$(basename "$key")
 
               if [ -z "${ALLOWED[$filename]+x}" ]; then
                 echo "Deleting unsupported archive: ${key} (${size} bytes)"
-                aws s3 rm "s3://${BUCKET}/${key}" || echo "::warning::Failed to delete ${key}"
-                echo "$filename" >> old_files.txt
-                DELETED_COUNT=$((DELETED_COUNT + 1))
+                if aws s3 rm "s3://${BUCKET}/${key}"; then
+                  echo "$filename" >> old_files.txt
+                  DELETED_COUNT=$((DELETED_COUNT + 1))
 
-                # Also delete the corresponding .log file if it exists
-                log_key="${key%.zst}.log"
-                aws s3 rm "s3://${BUCKET}/${log_key}" 2>/dev/null || true
+                  # Also delete the corresponding .log file if it exists
+                  log_key="${key%.zst}.log"
+                  aws s3 rm "s3://${BUCKET}/${log_key}" 2>/dev/null || echo "::warning::Failed to delete log ${log_key}"
+                else
+                  echo "::warning::Failed to delete ${key}"
+                fi
               else
                 echo "Keeping: ${key}"
                 KEPT_COUNT=$((KEPT_COUNT + 1))
--- a/.github/workflows/purge_debug_offline_archives.yml
+++ b/.github/workflows/purge_debug_offline_archives.yml
@@ -53,14 +53,14 @@ jobs:
           echo "Dry run:       ${DRY_RUN}"
           echo ""
 
-          # List all objects under the debug prefix
+          # List all objects under the debug prefix (with --output text, empty string is returned when no objects)
           OBJECTS=$(aws s3api list-objects-v2 \
             --bucket "${BUCKET}" \
             --prefix "${PREFIX}" \
             --query 'Contents[*].[Key,LastModified,Size]' \
-            --output text 2>/dev/null || true)
+            --output text 2>&1) || { echo "::warning::Failed to list S3 objects. Check bucket=${BUCKET} prefix=${PREFIX} and AWS credentials."; exit 0; }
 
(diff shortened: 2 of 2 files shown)


.github/workflows/purge_debug_offline_archives.yml
```diff
@@ -0,0 +1,134 @@
...
+            FILE_EPOCH=$(date -d "${last_modified}" +%s)
+
+            if [ "${FILE_EPOCH}" -lt "${CUTOFF_EPOCH}" ]; then
+              AGE_HOURS=$(( ($(date +%s) - FILE_EPOCH) / 3600 ))
+
+              if [ "${DRY_RUN}" = "true" ]; then
+                echo "[DRY RUN] Would delete: ${key} (age: ${AGE_HOURS}h, size: ${size} bytes)"
+              else
+                echo "Deleting: ${key} (age: ${AGE_HOURS}h, size: ${size} bytes)"
+                aws s3 rm "s3://${BUCKET}/${key}" || echo "::warning::Failed to delete ${key}"
```
Hahihula: if this fails, tha warning is logged but the DELETED_COUNT is still incremented

.github/workflows/purge_debug_offline_archives.yml
```diff
@@ -0,0 +1,134 @@
...
+          echo ""
+
+          # List all objects under the debug prefix
+          OBJECTS=$(aws s3api list-objects-v2 \
+            --bucket "${BUCKET}" \
+            --prefix "${PREFIX}" \
+            --query 'Contents[*].[Key,LastModified,Size]' \
+            --output text 2>/dev/null || true)
+
+          if [ -z "${OBJECTS}" ] || [ "${OBJECTS}" = "None" ]; then
```
Hahihula: with the --output text if no objects are there empty string is returned not "None"

.github/workflows/build_offline_installer_archives.yaml
```diff
@@ -412,29 +412,83 @@ jobs:
...
+          KEPT_COUNT=0
+
+          if [ -n "${S3_OBJECTS}" ] && [ "${S3_OBJECTS}" != "None" ]; then
+            while IFS=$'\t' read -r key size; do
+              [ -z "${key}" ] && continue
+              filename=$(basename "$key")
+
+              if [ -z "${ALLOWED[$filename]+x}" ]; then
+                echo "Deleting unsupported archive: ${key} (${size} bytes)"
+                aws s3 rm "s3://${BUCKET}/${key}" || echo "::warning::Failed to delete ${key}"
```
Hahihula: if this fails, warning is logged but delete_count is still incremented

.github/workflows/purge_debug_offline_archives.yml
```diff
@@ -0,0 +1,134 @@
...
+          echo "Cutoff time:   ${CUTOFF_HUMAN}"
+          echo "Dry run:       ${DRY_RUN}"
+          echo ""
+
+          # List all objects under the debug prefix
+          OBJECTS=$(aws s3api list-objects-v2 \
+            --bucket "${BUCKET}" \
+            --prefix "${PREFIX}" \
+            --query 'Contents[*].[Key,LastModified,Size]' \
+            --output text 2>/dev/null || true)
```
Hahihula: Maybe also print warning instead of silently ignoring the error?

.github/workflows/build_offline_installer_archives.yaml
```diff
@@ -412,29 +412,83 @@ jobs:
...
+          echo ""
+          echo "Total allowed archives: ${#ALLOWED[@]}"
+          echo ""
+
+          # List all .zst objects under dl/eim/archive_ (excludes debug/ since those are dl/eim/debug/archive_)
+          S3_OBJECTS=$(aws s3api list-objects-v2 \
+            --bucket "${BUCKET}" \
+            --prefix "${PREFIX}" \
+            --query 'Contents[?ends_with(Key, `.zst`)].[Key,Size]' \
+            --output text 2>/dev/null || true)
```
Hahihula: maybe log a warning/error instead of silently failing ?

.github/workflows/purge_debug_offline_archives.yml
alirana01: fixed now

.github/workflows/purge_debug_offline_archives.yml
alirana01: fixed

.github/workflows/build_offline_installer_archives.yaml
alirana01: fixed

.github/workflows/purge_debug_offline_archives.yml
alirana01: fixed

.github/workflows/build_offline_installer_archives.yaml
alirana01: fixed

Hahihula (approved):
Thanks for the changes. LGTM. GJ! And cross fingers on the first run :-)

PR #426, opened 2026-02-12
UI: Polish Chinese UI text

Shen7436:
This PR refines several Chinese UI expressions to improve clarity and avoid user confusion. 
See details in [DOC-13809](https://jira.espressif.com:8443/browse/DOC-13809).

### DIFF
--- a/src/locales/cn.json
+++ b/src/locales/cn.json
@@ -200,7 +200,7 @@
       },
       "custom": {
         "title": "自定义安装",
-        "description": "完全控制安装设置，适用于高级用户。",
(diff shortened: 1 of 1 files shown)


src/locales/cn.json
Shen7436, line 521: Hi @Hahihula The spell checker found a typo in the key name `prerequisitiesCheck` (should be `prerequisitesCheck`).

Would you like me to fix it across the codebase in this PR, or should we leave it as-is and add it to the spell checker's ignore list (not sure if this is feasible) for now?

PR #428, opened 2026-02-12
fix(linux): prevent WebKitGTK DMA-BUF crash on Nvidia + Wayland (EIM-500)

### LINKED ISSUE #421: EIM GUI crashes on Wayland + Nvidia (GBM buffer allocation failure) (EIM-496)
## Description

EIM GUI fails to launch on Linux systems with **Nvidia GPU + Wayland** due to a WebKitGTK DMA-BUF renderer issue. The app crashes immediately with:

```
Running GUI...
Failed to create GBM buffer of size 1200x1000: Invalid argument
```

Or when using `GDK_BACKEND=x11`:

```
Gdk-Message: Error 71 (Protocol error) dispatching to Wayland display.
```

## Environment

- **OS**: Fedora 43 (Wayland session)
- **GPU**: Nvidia (proprietary driver)
- **EIM version**: 0.7.1
- **Desktop**: GNOME on Wayland

## Root Cause

WebKitGTK's DMA-BUF renderer attempts to use GBM (Generic Buffer Management) for GPU buffer allocation. Nvidia's GBM implementation doesn't support all the formats/modifiers that WebKitGTK requests, causing the buffer allocation to fail and the app to crash.

This is a known issue across Tauri/WebKitGTK applications on Nvidia + Wayland:

- https://github.com/tauri-apps/tauri/issues/13493
- https://github.com/tauri-apps/tauri/issues/8254
- https://github.com/tauri-apps/tauri/issues/8254#issuecomment-3794663161
- https://github.com/TwintailTeam/TwintailLauncher/issues/36
- https://github.com/winfunc/opcode/issues/26
- https://github.com/DioxusLabs/dioxus/issues/1909

## Workaround

Setting `WEBKIT_DISABLE_DMABUF_RENDERER=1` forces WebKitGTK to use shared memory buffers instead of GBM, which works reliably:

```bash
WEBKIT_DISABLE_DMABUF_RENDERER=1 eim
```

## Suggested Fix

EIM could detect Nvidia + Wayland at startup and automatically set `WEBKIT_DISABLE_DMABUF_RENDERER=1`. For example in the launcher/main:

```rust
// Detect Nvidia GPU on Wayland and disable DMA-BUF renderer
if std::env::var("WAYLAND_DISPLAY").is_ok() || std::env::var("XDG_SESSION_TYPE").map(|v| v == "wayland").unwrap_or(false) {
    // Check for Nvidia GPU via lspci or /proc/driver/nvidia
    if std::path::Path::new("/proc/driver/nvidia").exists() {
        std::env::set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
    }
}
```

Alternatively, the `.desktop` file could include the env var by default, or the RPM packaging could set it conditionally.

## Related

- #379 — Fedora package manager detection bug (separate issue, same platform)

OmerFarukOruc:
## Summary

Fixes #421 — EIM GUI crashes immediately on Linux systems with Nvidia proprietary drivers running a Wayland session due to WebKitGTK's DMA-BUF renderer failing GBM buffer allocation.

## Changes

Added runtime detection in `src-tauri/src/gui/mod.rs` → `run()` that sets `WEBKIT_DISABLE_DMABUF_RENDERER=1` when:

1. **Wayland session detected** — via `XDG_SESSION_TYPE=wayland` or `WAYLAND_DISPLAY` being set
2. **Nvidia proprietary driver loaded** — via `/proc/driver/nvidia/version` existence

This follows the same pattern as the existing macOS PATH workaround directly above it.

### Design decisions

- **Respects user override**: Skips entirely if `WEBKIT_DISABLE_DMABUF_RENDERER` is already set, so users can opt out
- **`#[cfg(target_os = "linux")]`**: Compiled out on other platforms — zero cost
- **Placement**: Before `setup_gui_logging()` and `tauri::Builder`, ensuring the env var is set before WebKitGTK initialization
- **`eq_ignore_ascii_case`**: Handles potential casing variations without allocation

### Detection rationale

| Check | Why |
|-------|-----|
| `XDG_SESSION_TYPE` | Primary Wayland indicator set by display managers |
| `WAYLAND_DISPLAY` | Fallback — always present in Wayland sessions |
| `/proc/driver/nvidia/version` | Exists only when Nvidia's proprietary kernel module is loaded |

## Testing

Verified the workaround resolves the crash on:
- Arch Linux, Nvidia 570.133.07, Wayland (GNOME), kernel 6.13.4

The env var `WEBKIT_DISABLE_DMABUF_RENDERER` is a well-known workaround used across WebKitGTK-based applications (Tauri, Electron via similar mechanisms).

### DIFF
--- a/src-tauri/src/gui/mod.rs
+++ b/src-tauri/src/gui/mod.rs
@@ -238,6 +238,8 @@ pub fn run(log_level_override: Option<log::LevelFilter>) {
         env::set_var("PATH", "/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/local/bin:/opt/local/sbin");
     }
 
+    let _ = setup_gui_logging(log_level_override);
+
     // Workaround for WebKitGTK DMA-BUF renderer crash on Nvidia + Wayland (#421).
     // The GBM buffer allocation fails when using the DMA-BUF renderer with
     // Nvidia's proprietary driver on Wayland, causing the app to crash.
@@ -253,11 +255,10 @@ pub fn run(log_level_override: Option<log::LevelFilter>) {
 
             if is_wayland && has_nvidia {
                 env::set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
+                info!("Nvidia + Wayland detected: disabled WebKitGTK DMA-BUF renderer (WEBKIT_DISABLE_DMABUF_RENDERER=1)");
             }
         }
     }
-
-    let _ = setup_gui_logging(log_level_override);
 
     tauri::Builder::default()
         .plugin(tauri_plugin_dialog::init())


src-tauri/src/gui/mod.rs
```diff
@@ -238,6 +238,25 @@ pub fn run(log_level_override: Option<log::LevelFilter>) {
...
+        if env::var_os("WEBKIT_DISABLE_DMABUF_RENDERER").is_none() {
+            let is_wayland = env::var("XDG_SESSION_TYPE")
+                .map(|v| v.eq_ignore_ascii_case("wayland"))
+                .unwrap_or(false)
+                || env::var_os("WAYLAND_DISPLAY").is_some();
+
+            let has_nvidia = Path::new("/proc/driver/nvidia/version").exists();
+
+            if is_wayland && has_nvidia {
+                env::set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
```
alirana01, line 257: @OmerFarukOruc thanks a lot for the PR looks good maybe just a small log line to tell us when this is detected and the workaround is applied would be nice for debug and tracking in future.

OmerFarukOruc:
Thanks for the review! Good call — added an `info!()` log line when the workaround kicks in. Also moved the detection block after logging init so the message actually gets captured. See f35bc72.

PR #452, opened 2026-02-18
EIM-512 switched the output from the run command to streamig

Hahihula:
this should improve the experience from the run command. It will no longer wait for the command to finish but stream the output directly.


Fabricio-ESP:
Thanks Petr for the update.
Confirmed working as expected on MacOS.

Issue on Windows. Using powershell there is no return printed on terminal, neither streamed to terminal, or after the job is completed.

Hahihula:
> Thanks Petr for the update. Confirmed working as expected on MacOS.
> 
> Issue on Windows. Using powershell there is no return printed on terminal, neither streamed to terminal, or after the job is completed.



> Thanks Petr for the update. Confirmed working as expected on MacOS.
> 
> Issue on Windows. Using powershell there is no return printed on terminal, neither streamed to terminal, or after the job is completed.

fixed the windows issue:
image (1)

PR #468, opened 2026-02-24
EIM-535 enabled selecting idf based on idf path

Hahihula:
currently eim select command consumed only id or name of the idf or interactive selection, now it also accepts idf path. This will enable us to call it easily from export.sh script


alirana01 (approved):
LGTM!

alirana01:
@Hahihula just a small thing the pr link sync with jira works if you add a `:` after the ticket like `EIM-535: ******`

PR #469, opened 2026-02-24
EIM-536 fixed the eim remove and eim purge command for case the eim is installed from existing local repository or imported

Hahihula:
as @kolipakakondal reported, there is issue with the remove or purge command on eim in case idf was either imported or installed from local repository. After this change the eim will never remove non-empty directory so it can not remove some existing project.


alirana01 (approved):
LGTM!

PR #471, opened 2026-02-24
EIM-474: python app execution alias detection windows

alirana01:
## Summary

**Jira:** [EIM-474](https://jira.espressif.com/browse/EIM-474)
On Windows, `python.exe` and `python3.exe` on PATH can point to **App Execution Aliases** — stubs under `%LOCALAPPDATA%\Microsoft\WindowsApps` that redirect to the Microsoft Store and break IDF tooling. This PR adds detection and user guidance so we skip aliases when resolving Python and fail the sanity check with a clear message when only the alias is available.

## What changed

- **Alias detection** — New Windows-only `is_app_execution_alias(path)` in `utils.rs` (reparse point with `IO_REPARSE_TAG_APPEXECLINK`). Winapi features `fileapi` and `minwinbase` added in `Cargo.toml`.
- **Python resolution** — On Windows, `detect_default_python()` now uses `where.exe` and filters out alias paths via `resolve_python_path()` so the chosen interpreter is a real executable. Scoop shims are still tried first; then PATH candidates are resolved while skipping aliases and verified with `is_python3()`.
- **New sanity check** — **App Execution Alias** (Windows only), run first in the Python sanity check. If the resolved Python is an alias, the check fails with a message pointing users to install from [python.org](https://www.python.org/downloads/) or disable the alias in **Settings → Apps → Advanced app settings → App execution aliases**. Localized (EN/CN) in `locales/app.yml`.
- **No programmatic disable** — We do not turn off aliases; we only detect them and guide the user.

## Files changed

| File | Change |
|------|--------|
| `src-tauri/Cargo.toml` | Winapi features `fileapi`, `minwinbase` |
| `src-tauri/locales/app.yml` | App Execution Alias check + hint (en/cn) |
| `src-tauri/src/lib/python_utils.rs` | Alias-aware resolution, new sanity check, integrated with unified `detect_default_python()` |
| `src-tauri/src/lib/utils.rs` | `is_app_execution_alias(path)` (Windows) |

## How to test (Windows)

1. **Alias only** — With only the App Execution Alias on PATH, run the Python sanity check → expect **App Execution Alias** check to **fail** with a clear hint.
2. **Real Python on PATH** — With python.org Python on PATH (alias can still be present), run the sanity check → expect **App Execution Alias** to **pass** and remaining checks to use the real Python.
3. **Explicit path** — Set Python to the alias path in the app → check fails; set to a real interpreter path → check passes.
4. **Venv creation** — With real Python on PATH (no Scoop), run a flow that creates a venv → venv should be created with the real Python, not the alias.

Detailed steps: see `out/agent-temp/eim-474-pr-summary/EIM-474-summary-and-windows-testing.md` in the repo.

## Checklist

- [x] Rebased on latest `master`
- [ ] Sanity check fails with clear message when only App Execution Alias is available (Windows)
- [ ] Sanity check passes when real Python is on PATH (Windows)
- [ ] Venv creation uses real Python when Scoop is not used (Windows)
- [ ] No regression on non-Windows platforms

### DIFF
--- a/src-tauri/src/lib/python_utils.rs
+++ b/src-tauri/src/lib/python_utils.rs
@@ -1051,10 +1051,12 @@ fn detect_default_python() -> String {
             "python".to_string()
         }
         _ => {
+            // Unix: prefer python3
             if is_python3("python3") {
                 info!("Found python3");
                 return "python3".to_string();
             }
+            // Fallback: bare `python`, but only if it's actually 3.x
             if is_python3("python") {
                 info!("Found python (verified 3.x)");
                 return "python".to_string();


src-tauri/src/lib/python_utils.rs
```diff
@@ -959,60 +993,122 @@ fn check_python_version(py: &str) -> GenericCheckResult<SanityCheck> {
...
-            if let Some(scoop_shims) = get_scoop_path() {
-                let scoop_dir = PathBuf::from(&scoop_shims);
-                for name in &["python3.exe", "python.exe"] {
-                    let candidate = scoop_dir.join(name);
-                    let candidate_str = candidate.to_string_lossy();
-                    if is_python3(&candidate_str) {
-                        info!("Using Scoop {}: {}", name, candidate.display());
-                        return candidate_str.into_owned();
-                    }
+    #[cfg(windows)]
```
Hahihula: This place formely used
`match std::env::consts::OS {`
instead of build macros `#[cfg(windows)]` 
as it's done throughout the whole project.
Can you explain to me why did you decided to change that here?

src-tauri/src/lib/python_utils.rs
```diff
@@ -959,60 +993,122 @@ fn check_python_version(py: &str) -> GenericCheckResult<SanityCheck> {
...
+        }
+        return Some(candidate);
+    }
+    None
+}
+
+/// Fallback for Windows when Scoop is not available: tries to find a real
+/// python3.exe or python.exe via `resolve_python_path`, skipping aliases.
+#[cfg(windows)]
+fn resolve_python_fallback_windows() -> String {
```
Hahihula: this function seems to be completely unused

src-tauri/src/lib/python_utils.rs
```diff
@@ -959,60 +993,122 @@ fn check_python_version(py: &str) -> GenericCheckResult<SanityCheck> {
...
 }
 
+/// Resolves an executable name (e.g. `"python"`) to the first real path on
+/// the system PATH that is **not** a Windows App Execution Alias.
+///
+/// Uses `where.exe` to enumerate all candidates, then filters out any that
+/// are reparse points with `IO_REPARSE_TAG_APPEXECLINK`.  Returns `None`
+/// when every candidate is an alias or `where` finds nothing.
+#[cfg(windows)]
+fn resolve_python_path(name: &str) -> Option<PathBuf> {
```
Hahihula: we have the detect_default_python you implemented for the sanity check
this seems to be duplicated logic

src-tauri/src/lib/python_utils.rs
```diff
@@ -60,6 +62,11 @@ impl SanityCheck {
         };
         format!("python.sanitycheck.hint.{}{}", self.key_name(), os_suffix)
     }
+
+    /// Returns `true` for checks that only apply on Windows.
+    pub fn is_windows_only(&self) -> bool {
```
Hahihula: is this ever used?

Hahihula (commented):
care to explain the change of the approcach from match std::env::consts::OS { to macros?
also this PR seems to have some unused code.

PR #472, opened 2026-02-24
EIM-538 add ci pipeline for arm7 builds

Hahihula:


.github/workflows/build.yaml
```diff
@@ -350,6 +350,238 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: armv7-unknown-linux-musleabihf
+
+      - name: Install cross
+        run: cargo install cross --git https://github.com/cross-rs/cross
```
alirana01: This is probably still a wip but shouldnt we be using a tag for the specific version as this will lead to using default branch that may not be stable

.github/workflows/build.yaml
alirana01, line 1: `build-cli-linux-armv7` has `needs: [setup, build-test-lib]` and runs in parallel with `build-cli-linux`. The step “Download completions from linux-x64 build” downloads the artifact `eim-cli-linux-x64-...`, which is produced by the `build-cli-linux` job. If the ARMv7 job runs or reaches that step before `build-cli-linux` has finished and uploaded the artifact, the download can fail.

Hahihula: good catch, mised that, fixed

Issue #475, opened 2026-02-24
EIM and EIDE disagree on venv python version (EIM-541)


gtjoseph:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

You should be able to build a project from the EIDE and CLI without having to reconfigure and rebuild when switching between the two.

### Actual behavior (suspected bug)

Building from the CLI after building from EIDE causes errors.

### Error logs or terminal output

```shell
$ idf.py build
Executing action: all (aliases: build)
'/home/gtjoseph/.espressif/tools/python/v5.5.3/venv/bin/python3' is currently active
 in the environment while the project was configured with 
'/home/gtjoseph/.espressif/tools/python/v5.5.3/venv/bin/python'.
Run 'idf.py fullclean' to start again.
```

### Steps to reproduce the behavior

1. Build a project from the EIDE
2. From the CLI, source the "activate" script from ~/.espressif/tools that corresponds to the ESP-IDF version configured in EIDE.
3. Build from the CLI with `idf.,py build`

### Project release version

EIM 0.8.0 EIDE 4.1.0

### System architecture

Intel/AMD 64-bit (modern PC, older Mac)

### Operating system

Linux

### Operating system version

Fedora 43

### Shell

ZSH

### Additional context

The EIM creates an activate script in `~/.espressif/tools` for each version of the ESP-IDF installed.  In those scripts, it uses "python3" as the python executable...

```shell
alias idf.py="/home/gtjoseph/.espressif/tools/python/87912cd/venv/bin/python3 /usr/src/mcu/esp32/esp-idf/tools/idf.py"
```

The EIDE however, references the python executable as just "python"
```
-DPYTHON=/home/gtjoseph/.espressif/tools/python/87912cd/venv/bin/python
```

This causes issues if you build from EIDE then attempt a build from the CLI even though both `python` and `python3` are symlinks to `python3.14` :

```
$ idf.py build
Executing action: all (aliases: build)
'/home/gtjoseph/.espressif/tools/python/v5.5.3/venv/bin/python3' is currently active in the environment while the project was configured with '/home/gtjoseph/.espressif/tools/python/v5.5.3/venv/bin/python'. Run 'idf.py fullclean' to start again.
```

[...]

PR #481, opened 2026-02-25
EIM-541: Fix for venv python alias in export

alirana01:
User reported bug #475 where the alias in activation scripts was pointing to python3 causing a mismatch with ide and cli.

For testing please use the original reported bug


Hahihula:
https://github.com/espressif/idf-im-ui/pull/489

PR #486, opened 2026-02-26
fixed Os name detection on linux

Hahihula:
image

### DIFF
--- a/src-tauri/src/gui/commands/utils_commands.rs
+++ b/src-tauri/src/gui/commands/utils_commands.rs
@@ -1,5 +1,5 @@
 use idf_im_lib::{self, ensure_path};
-use idf_im_lib::telemetry::track_event;
+use idf_im_lib::telemetry::{get_linux_os_name, track_event};
 use log::{error, info};
 use serde_json::{json,Value};
 use tauri_plugin_store::StoreExt;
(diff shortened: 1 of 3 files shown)


src-tauri/src/gui/commands/utils_commands.rs
```diff
@@ -58,7 +58,9 @@ pub fn get_operating_system() -> String {
     let mut sys = System::new_all();
     sys.refresh_all();
 
-    System::name().unwrap_or_else(|| std::env::consts::OS.to_string())
+    System::name()
+        .filter(|s| !s.is_empty() && s != "Unknown")
+        .unwrap_or_else(|| get_linux_os_name())
```
alirana01: Shouldnt this be something like this 

```rust
{
            if std::env::consts::OS == "linux" {
                idf_im_lib::telemetry::get_linux_os_name()
            } else {
                std::env::consts::OS.to_string()
            }
        }
```

src-tauri/src/lib/telemetry.rs
```diff
@@ -75,18 +75,30 @@ fn parse_connection_string() -> (String, String) {
...
+        if let Some(name) = line.strip_prefix("ID=") {
+          let distro = name.trim_matches('"').to_lowercase();
+          return format!("linux-{}", distro);
+        }
+      }
+    }
+    // Fallback to generic linux if ID not found
+    return "linux".to_string();
+  }
+  "Unknown".to_string()
```
alirana01, line 107: I dont know this is doing what it should be somehow I think since i can understand it with comparison to my java this may not be a very good rust code we so many nested ifs and early returns shouldn't we be using match pattern here to ease out early returns?

PR #538, opened 2026-03-05
EIM-589 fixed get git path

Hahihula:


Fabricio-ESP (approved):
Thanks for the quick change.

PR #550, opened 2026-03-07
EIM-595: CLI Tests for using existing cloned git repo to verify the installation

alirana01:
## Summary

Adds an **end-to-end CLI test** that verifies EIM can install tools and environment for an **already-cloned** ESP-IDF git repository: the suite clones `esp-idf`, runs `eim install -p <path>` (path quoted on Windows), then runs the shared **installation verification** pipeline (`eim_idf.json`, activation, Python checks, sample project / build where applicable, and **`idf.py --version`**).

The scenario matches a user who has already cloned `esp-idf` and wants EIM to provision tools for that tree **without** re-cloning via EIM.

**Windows Scoop CI:** The branch **splits** the default CLI suite from the Scoop matrix: **stock Windows** runs the full suite (including this test); **Scoop** runs a reduced suite **without** the existing-clone test, because the Scoop job strips Git from `PATH` before tests while `CLICloneIDFRepo` needs **git** to clone. See *CI & Scoop separation* below.

---

### 1. Suite entry — `existing-git-clone`

- **File:** `tests/suites/CLI-basic.json`
- **New entry:** `type: "existing-git-clone"` (id `6`), name *Install over existing git cloned repo folder*.
- **Data:** `installFolder` (under the user home, e.g. `.espext1`), `gitRepoUrl`, `gitRepoBranch` (default `v5.5.3`), plus standard flags (`deleteAfterTest`, `testProxyMode`).

There is **no** separate `CLI-existing-repo-cloned-git.json` file on this branch: the flow is **one suite entry inside `CLI-basic.json`**.

### 2. Runner wiring — `tests/CLIRunner.test.js`

For `test.type === "existing-git-clone"`, the runner chains:

1. **`runCLIClonedIDFRepo`** (`tests/scripts/CLICloneIDFRepo.test.js`) — clones the repo to the configured path and asserts the clone layout.
2. **`runCLICustomInstallTest`** (`tests/scripts/CLICustomInstall.test.js`) — non-interactive install with `eim install -p <path>` (quoted on Windows).
3. **`runInstallVerification`** (`tests/scripts/installationVerification.test.js`) with **`existingGitClone: true`** — resolves the `eim_idf.json` entry by **resolved clone path**, validates metadata and paths, and runs the usual verification steps (including `idf.py --version`).
4. **`runCleanUp`** — teardown when `deleteAfterTest` is true.

This reuses existing install and verification code paths instead of a single standalone test module.

### 3. Clone helper

- **File:** `tests/scripts/CLICloneIDFRepo.test.js`
- **Export:** `runCLIClonedIDFRepo({ id, path, gitRepoUrl, gitRepoBranch })` — `git clone` to the target path; asserts the folder exists and looks like IDF (`tools/`, etc.).

### 4. Verification behavior for existing clones — `installationVerification.test.js`

- **`existingGitClone: true`** switches entry lookup to **`path.resolve()`** matching the install folder, and expects the IDF path in JSON to match the clone root.
- **Sample project layout:** Projects are validated **beside** the clone (sibling `projects/` directory), **not** under `esp-idf`, so verification does not create trees under the clone that would dirty the user’s git repo or leave empty dirs after `eim remove` (see `getProjectsDir()` and commit `600a135`).
- **`idf.py --version`:** Test **8** still checks that output includes the expected tag/branch and, **in current code**, that output **does not** contain `-dirty`. If you want to allow dirty working trees for real user clones, consider relaxing that assertion when `existingGitClone` is true (not done on this branch).

### 5. CI & Scoop separation

| Platform / matrix | Suite file | Notes |
|-------------------|------------|--------|
| Linux, macOS, Windows **stock** | `CLI-basic` | Includes **`existing-git-clone`** with other CLI tests. |
| Windows **scoop** | `CLI-scoop-windows` | Same entries as `CLI-basic` **except** the `existing-git-clone` block is omitted. |

### DIFF
--- a/tests/scripts/installationVerification.test.js
+++ b/tests/scripts/installationVerification.test.js
@@ -40,6 +40,17 @@ export function runInstallVerification({
       return path.join(installFolder, idf, "esp-idf");
     }
 
+    /**
+     * Sample projects live beside the esp-idf tree, not under it (avoids dirty git, empty dirs after eim remove).
+     * Default: installFolder/idf/projects. Existing-clone: sibling projects/ next to the clone root.
+     */
+    function getProjectsDir(idf) {
+      if (existingGitClone) {
+        return path.join(path.dirname(installFolderResolved), "projects");
+      }
+      return path.join(installFolder, idf, "projects");
+    }
+
     /** Activation script path for default layout; when existingGitClone use entry.activationScript. */
     function getActivationScriptPath(idf, entry) {
       if (existingGitClone && entry) return entry.activationScript;
@@ -475,7 +486,8 @@ export function runInstallVerification({
         expect(eimJsonEntry, `No entry for IDF ${idf} in eim_idf.json`).to.not.be.null;
 
         testRunner = new CLITestRunner();
-        const pathToProjectFolder = path.join(getIdfRoot(idf), "project");
+        const projectsDir = getProjectsDir(idf);
+        const pathToProjectFolder = projectsDir;
         const activationScript = getActivationScriptPath(idf, eimJsonEntry);
         try {
           await testRunner.runIDFTerminal(activationScript);
@@ -552,9 +564,8 @@ export function runInstallVerification({
         expect(eimJsonEntry, `No entry for IDF ${idf} in eim_idf.json`).to.not.be.null;
 
         testRunner = new CLITestRunner();
-        let pathToProjectFolder = path.join(
-          getIdfRoot(idf),
-          "project",
+        const pathToProjectFolder = path.join(
+          getProjectsDir(idf),
           "hello_world"
         );
         const activationScript = getActivationScriptPath(idf, eimJsonEntry);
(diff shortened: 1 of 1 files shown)


tests/scripts/installationVerification.test.js
```diff
@@ -451,19 +463,18 @@ export function runInstallVerification({
...
-        let pathToProjectFolder = path.join(installFolder, idf, "project");
-        const pathToIDFScript =
-          os.platform() !== "win32"
-            ? path.join(toolsFolder, "tools", `activate_idf_${idf}.sh`)
-            : path.join(
-                toolsFolder,
-                "tools",
-                `Microsoft.${idf}.PowerShell_profile.ps1`
-              );
+        const pathToProjectFolder = path.join(getIdfRoot(idf), "project");
```
Fabricio-ESP: On this refactoring, you are creating the project folder inside the IDF folder itself. This have 2 effects:
The new project folder sits in .espressif/v5.5.3/esp-idf/projects
But itis supposed to be on .espressif/v5.5.3/projects (outside the git folder)

The first effect is that the git folder will not be --dirty as there are new files modified inside the git folder

When we remove the IDF installation it removes the "esp-idf" and the part folder IF there isn`t any file or folder. This is causing a test failure because the projects folder is not in the wrong place and there is nothing else insde the v5.5.3 folder, so everything is deleted.

It is better to keep the projects folder outside the git folder.

tests/scripts/installationVerification.test.js
Fabricio-ESP, line 590: The set target command when git folder is previously cloned failed on one of the runners here. This is because the set-target command will trigger the submodules upgrade, which takes forever when github is busy. You might need to extend the idle timeout time, because the submodules clone does not add much data to the stdout while downloading it.

PR #556, opened 2026-03-11
EIM-600: Offline archives symlinks and extraction respecting original archive file params

alirana01:
## Problem

After installing ESP-IDF from an **offline archive**, running `git status` inside the installed IDF directory showed a **dirty working tree** (unintended changes). This did not happen when installing the same version via the normal (online) flow. The dirty state made it look as if files had been modified and could confuse users and tooling that rely on a clean repo.

## Root causes

Two separate issues contributed to the dirty status:

### 1. Symlinks were stored as regular files in the archive

When building the offline `.tar.zst` archive, the `tar-rs` crate was used with its **default** behavior of **following symlinks**. That meant:

- Symlinks in the cloned IDF repo (e.g. scripts under `tools/`) were **not** stored as symlink entries in the tar.
- Instead, the **content of the symlink targets** was archived as normal files.
- On extraction, those entries became regular files rather than symlinks.
- Git then saw a difference: index had symlinks, working tree had regular files → **dirty status**.

### 2. File modification times were lost during copy

When moving/copying the extracted archive contents to the final install location:

- **Windows:** The previous approach used PowerShell `cp -r`, which did not reliably preserve timestamps and had long-path limitations.
- **Non-Windows:** A plain copy was used instead of a move where possible, and the copy did not preserve modification times (mtime).
- Git uses mtime (among other things) to detect changes. When mtime was reset, git could report files as modified → **dirty status**.

### Archive creation (offline installer builder)

- **`src-tauri/src/offline_installer_builder.rs`**  
  Before adding the archive directory to the tar, call `tar.follow_symlinks(false)`.  
  Symlinks are now stored as **symlink entries** in the archive, so extraction restores them correctly and git no longer sees symlinks turned into regular files.

### Extraction and copy to install path

- **`src-tauri/src/lib/offline_installer.rs`** — `copy_idf_from_offline_archive`:
  - **Windows:** Use **robocopy** instead of PowerShell `cp -r`:
    - Handles long paths (avoids MAX_PATH issues).
    - Uses `/COPY:DAT` to copy Data, Attributes, and Timestamps, so mtime is preserved.
    - Exit codes 0–7 are treated as success; failures are logged and reported.
  - **Non-Windows:** Prefer **`fs::rename`** (move) when possible so metadata is preserved. If `rename` fails (e.g. cross-filesystem), fall back to a **mtime-preserving copy** instead of a plain copy.

- **`src-tauri/src/lib/utils.rs`** — new helper:
  - **`copy_dir_contents_preserving_mtime`** recursively copies a directory and, for each file, copies the content then sets the destination file’s modification time to match the source (using the `filetime` crate). This keeps git’s view of the working tree consistent when a move is not possible.

- **Path handling:** Use `to_string_lossy().as_ref()` for paths passed to `robocopy` so that paths with non-UTF-8 characters (e.g. some Windows user names) do not cause a panic; invalid sequences are replaced and robocopy failure is handled in the existing error path.

### Dependencies

- **`src-tauri/Cargo.toml`**  
  Added the **`filetime`** crate for setting file modification times in `copy_dir_contents_preserving_mtime`.

## Summary of changes

| Area | Change |
|------|--------|
| **Archive creation** | `tar.follow_symlinks(false)` so symlinks are stored as symlinks. |
| **Windows copy** | Use robocopy with `/COPY:DAT` for long paths and timestamp preservation; safe path strings via `to_string_lossy()`. |
| **Non-Windows copy** | Prefer `fs::rename`; fallback to `copy_dir_contents_preserving_mtime` when rename fails. |
| **New utility** | `copy_dir_contents_preserving_mtime` in `utils.rs` (uses `filetime`). |
| **Dependency** | Added `filetime` in `Cargo.toml` / `Cargo.lock`. |

With these changes, installing IDF from an offline archive results in a **clean `git status`** in the installed IDF directory, consistent with an online installation.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1417,6 +1417,7 @@ dependencies = [
  "dialoguer",
  "dirs 6.0.0",
  "fern",
+ "filetime",
  "flate2",
  "fork",
  "fs_extra",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -113,6 +113,7 @@ anyhow = "^1.0"
 rust_search = "2.1.0"
 flate2 = { version = "1.0", default-features = false, features = ["rust_backend"] }
 tar = { version = "0.4", default-features = false }
+filetime = "0.2"
 zip = { version = "8.1", default-features = true }
 lzma-rs = "0.3.0"
 thiserror = "2.0"
--- a/src-tauri/src/lib/git_tools.rs
+++ b/src-tauri/src/lib/git_tools.rs
@@ -1157,7 +1157,7 @@ fn checkout_submodule_worktree(
     // Need this to prevent untracked files in the submodule workdir
     populate_index_from_tree(&repo, &tree)?;
 
-    
+
     Ok(())
 }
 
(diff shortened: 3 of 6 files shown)


Hahihula (approved):
LGTM!

ekomarov128 (approved):

PR #582, opened 2026-03-20
Replaced the action we used for the github release assets upload with direct calling of the gh cli util

Hahihula:


kolipakakondal (approved):
Thanks for the fix. LGTM.

Issue #462, opened 2026-02-22
Installation on Fedora 43 fails to detect suitable python version (EIM-529)


dveeden:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

Installing on Fedora 43 fails due to the dependency on Python.

However I have both `python3` installed that gives me Python 3.14 and `python3.13` that gives me Python 3.13.

So it should use the older Python release if it isn't compatible with 3.14 (which other issues report about).

### Actual behavior (suspected bug)

Fails due to broken dependencies

### Error logs or terminal output

```shell
dvaneeden@dve-carbon:~$ cat /etc/yum.repos.d/espressif-eim.repo 
[eim]
name=ESP-IDF Installation Manager
baseurl=https://dl.espressif.com/dl/eim/rpm/$basearch
enabled=1
gpgcheck=0
dvaneeden@dve-carbon:~$ sudo dnf install eim
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: conflicting requests
  - nothing provides python3 < 3.14 needed by eim-0.8.0-1.x86_64 from eim
  - nothing provides python3 >= 3.10 needed by eim-0.8.0-1.x86_64 from eim
You can try to add to command line:
  --skip-broken to skip uninstallable packages
dvaneeden@dve-carbon:~$ . /etc/os-release
dvaneeden@dve-carbon:~$ echo $PRETTY_NAME
Fedora Linux 43 (Workstation Edition)
dvaneeden@dve-carbon:~$ rpm -qa | grep -E '^python3(\.|-[0-9])'
python3-3.14.2-1.fc43.x86_64
python3.13-libs-3.13.11-1.fc43.x86_64
python3.13-3.13.11-1.fc43.x86_64
dvaneeden@dve-carbon:~$ /usr/bin/python3 --version
Python 3.14.2
dvaneeden@dve-carbon:~$ /usr/bin/python3.13 --version
Python 3.13.11
```

### Steps to reproduce the behavior

Try to install on Fedora 43

### Project release version

0.8.0-1

### System architecture

Intel/AMD 64-bit (modern PC, older Mac)

### Operating system

Linux

### Operating system version

Fedora 43

### Shell

ZSH

[...]

PR #637, opened 2026-03-25
EIM-529: Move RPM build to Fedora image

Hahihula:
this should solve #462

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -295,68 +295,16 @@ jobs:
           cp "$DEB_FILE" "eim-cli-${{ matrix.package_name }}.deb"
           gh release upload "${{ github.event.release.tag_name }}" "eim-cli-${{ matrix.package_name }}.deb" --clobber
 
-      - name: Build CLI .rpm package
-        run: |
-          cd src-tauri
-          cargo install cargo-generate-rpm || true
-          if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
-            VERSION="${GITHUB_REF_NAME#v}"
-          else
-            VERSION="0.0.0-${GITHUB_SHA:0:7}"
-          fi
-          echo "Using version: $VERSION"
-          mkdir -p ~/.rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
-
-          cargo generate-rpm \
-            --target ${{ matrix.target || 'x86_64-unknown-linux-musl' }} \
-            --payload-compress gzip \
-            --set-metadata 'requires = {git = "*", wget = "*", flex = "*", cmake = "*", bison = "*", gperf = "*", ccache = "*", libffi-devel = "*", openssl-devel = "*", dfu-util = "*", libusb = "*"}'
-
-          # Find the generated RPM and add python3 dependency with correct RPM syntax
-          RPM_FILE=$(find target -name "*.rpm" -print -quit)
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -1285,6 +1294,181 @@ jobs:
           cp "$DMG_FILE" "eim-gui-${{ matrix.package_name }}.dmg"
           gh release upload "${{ github.event.release.tag_name }}" "eim-gui-${{ matrix.package_name }}.dmg" --clobber
 
+  # ============================================================
+  # Repackage GUI RPMs inside Fedora container
+  # ============================================================
+  build-gui-rpm:
```
alirana01, line 1300: we are missing a `actions/checkout@v4` for this `build-gui-rpm`

This can then cause failures when release assets are being uploaded in `Upload GUI RPM Release Asset` because `gh` will not have any local remote context for repo. So on release this will fail with not a git repository error on `Upload GUI RPM Release Asset` step

.github/workflows/build.yaml
```diff
@@ -1285,6 +1294,181 @@ jobs:
...
+
+      - name: Upload GUI RPM Release Asset
+        if: github.event_name == 'release' && github.event.action == 'created' && steps.find-gui-rpms.outputs.gui_rpm_files_found == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          RPM_FILE=$(ls gui_rpm_output/*.rpm | head -1)
+          cp "$RPM_FILE" "eim-gui-${{ matrix.package_name }}.rpm"
+          gh release upload "${{ github.event.release.tag_name }}" "eim-gui-${{ matrix.package_name }}.rpm" --clobber
+
```
alirana01, line 1471: affected place because of [this](https://github.com/espressif/idf-im-ui/pull/637/changes#r2993526132)

.github/workflows/build.yaml
```diff
@@ -532,31 +480,126 @@ jobs:
...
+            src-tauri/Cargo.toml
+            src-tauri/target/${{ env.CROSS_TARGET }}/release/eim
+            src-tauri/target/release/completions/
+
+  # ============================================================
+  # Build CLI RPM packages inside Fedora container
+  # ============================================================
+  build-cli-rpm:
+    name: Build CLI RPM (${{ matrix.package_name }})
+    needs: [build-cli-linux, build-cli-linux-armv7]
```
alirana01, line 499: One thing to adjust: `build-cli-rpm` now has `needs: [build-cli-linux, build-cli-linux-armv7]`, so an armv7 failure can block x64/aarch64 RPM publishing even when those artifacts are already produced.

Could we split this into two RPM jobs to keep per-arch isolation?

- `build-cli-rpm-main` (matrix: `linux-x64`, `linux-aarch64`) with `needs: [build-cli-linux]`
- `build-cli-rpm-armv7` (single entry: `linux-armv7`) with `needs: [build-cli-linux-armv7]`

Hahihula: As we need the support for all these architectures, I would prefer to fail as early as possible. Producing subset would be considered failure.

.github/workflows/build.yaml
Hahihula: Good point! Thanks for the catch.
i solved this issue by adding:
```
--repo "${{ github.repository }}" 
```
to the gh command

PR #692, opened 2026-04-13
UI: Clarify Python ssl check hints for network issues (EIM-708)

Carrot-shreds:
Lead users to check their network issues first when the python ssl check fails.
(I just spent hours checking and reinstalling my Python environment, and eventually found it was a problem with my proxy.)


Hahihula:
hi @Carrot-shreds,
I do not see any reason not to merge this, and as it may help people, I am going to do so.
Thanks for the contribution. 
We are trying to think of all the things that may happen, but there is no such thing as too clear instructions. 
Thanks.

Hahihula (approved):

PR #695, opened 2026-04-14
EIM-388: added aliases for esptoll espfuse and espsecure without the .py extension

Hahihula:


alirana01 (approved):
LGTM!
Just a small question shouldn't we add the aliases for other things like otatool as well?

PR #698, opened 2026-04-14
Eim 345: recreating config already installed idf

Hahihula:


src-tauri/src/lib/idf_config.rs
```diff
@@ -1,13 +1,105 @@
...
+                    .map_err(|e| DeError::custom(format!("Invalid base64: {}", e)))
+            }
+            Err(_) => {
+                // Legacy format: might be a sequence of integers
+                // Need to re-create deserializer since we consumed it
+                Err(DeError::custom("Expected base64 string or legacy integer array"))
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
alirana01: @Hahihula i think this makes sense

src-tauri/src/lib/idf_config.rs
```diff
@@ -1,13 +1,105 @@
...
+            for (i, v) in arr.into_iter().enumerate() {
+                if let serde_json::Value::Number(n) = v {
+                    match n.as_u64() {
+                        Some(n) => bytes.push(n as u8),
+                        None => return Err(DeError::custom(format!("Number at index {} is too large for u8", i))),
+                    }
+                } else {
+                    return Err(DeError::custom(format!("Expected number at index {}, got {:?}", i, v)));
+                }
+            }
```
alirana01: If i understood this correctly this integer array format will never exist in any real config so why do we need legacy here?

src-tauri/Cargo.toml
```diff
@@ -135,6 +135,8 @@ semver = "1.0.27"
 sysinfo = "0.38.0"
 libc = "0.2.180"
 bytes = "1.11.1"
+bincode = "1.3"
```
alirana01, line 138: not sure if we should be concerned that bincode is not maintained anymore but for our usecase as long as there are no issues we can use it but just a pointer

alirana01 (dismissed):
LGTM! Just a few small nitpicks that if you think are worth fixing and a small concern on bincode dep we added

src-tauri/Cargo.toml
Hahihula: it's not maintained because it's complete:
```
Version 1.3.3 is considered a complete version of bincode that is not in need of any updates. Updates will only be pushed to the in the unlikely event of CVEs.
```

src-tauri/src/lib/idf_config.rs
Hahihula: you are right, it;s residuum from the development phase and should not be there. Thanks for the catch.

tests/scripts/installationVerification.test.js
```diff
@@ -190,6 +190,20 @@ export function runInstallVerification({
...
+        // installationConfig is optional (for backwards compatibility with v1 configs)
+        // When present, it should be a base64 encoded string containing binary serialized Settings
+        if (idf.installationConfig !== undefined) {
+          expect(
+            idf.installationConfig,
+            `installationConfig should be a string (base64 encoded) for IDF ${idf.name || "invalid IDF Entry"}`
+          ).to.be.a("string");
+          // Basic base64 validation - should be valid base64 string
+          expect(
+            /^[A-Za-z0-9+/]*={0,2}$/.test(idf.installationConfig),
```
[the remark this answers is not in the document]
alirana01: @Hahihula I think this looks like a valid thing from Gemini not sure if the suggested regex is correct but if the idea is to expect Base64 and our regex is not perfect we can try to use a correct one

alirana01 (approved):
Thanks a small change maybe in Regex

Hahihula:
@alirana01 The regex is not meant to "validate corect encoding" but "validate not breaking json syntax"

PR #705, opened 2026-04-15
EIM-34 added retry logic to git operations

Hahihula:
should improve resiliency on poor network

### DIFF
--- a/src-tauri/src/lib/git_tools.rs
+++ b/src-tauri/src/lib/git_tools.rs
@@ -416,17 +416,22 @@ pub fn clone_repository(
         }
     }
 
-    // Fetch
     info!("Cloning repository from {:?}", options);
-    let (mut checkout, _) = match prepare
-        .fetch_then_checkout(gix::progress::Discard, should_interrupt){
-            Ok(res) => res,
-            Err(e) => {
-                let _ = tx.send(ProgressMessage::Finish);
-                error!("Failed to fetch repository: {}", e);
-                return Err(Box::new(e));
-            }
-        };
+    let mut prepare_mut = prepare;
+    let fetch_result = crate::utils::with_retry_exponential(
+        || prepare_mut.fetch_then_checkout(gix::progress::Discard, should_interrupt),
+        3, // max_retries
+        std::time::Duration::from_millis(500), // base_delay
+    );
+
+    let (mut checkout, _) = match fetch_result {
+        Ok(res) => res,
+        Err(e) => {
+            let _ = tx.send(ProgressMessage::Finish);
+            error!("Failed to fetch repository: {}", e);
+            return Err(Box::new(e));
+        }
+    };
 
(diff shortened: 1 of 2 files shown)


src-tauri/src/lib/utils.rs
```diff
@@ -524,12 +541,39 @@ where
...
-                debug!("Attempt {} failed with error: {:?}", attempt, e);
+                let delay = strategy.delay(attempt as u32);
+                match log_level {
+                    log::Level::Warn => warn!("Attempt {} failed: {:?}, retrying in {:?}", attempt, e, delay),
+                    log::Level::Debug => debug!("Attempt {} failed with error: {:?}", attempt, e),
+                    _ => {}
+                }
+                if delay > std::time::Duration::ZERO {
+                    std::thread::sleep(delay);
+                }
```
alirana01, line 552: I see some confusion coming from this, the attempt vs retries naming although seems to be consistent with what we have but they are not actually those its more like if the max_retries is 3 we do 3 attempts and 2 retries not important but a bit confusing to read.

src-tauri/src/lib/git_tools.rs
```diff
@@ -1098,20 +1103,30 @@ fn fetch_single_commit_to_modules(
...
+                .map_err(|e| format!("Failed to prepare fetch: {}", e))?
+                .with_shallow(shallow.clone())
+                .receive(gix::progress::Discard, &AtomicBool::new(false))
+                .map_err(|e| format!("Failed to receive: {}", e))
+        },
+        3,
+        std::time::Duration::from_millis(500),
+    ).map_err(|e| {
+        format!(
+            "fetch_single_commit_to_modules failed after 3 attempts: {}",
```
alirana01, line 1126: very small oversight but can make our lives easier later to actually log the attempts variable here

src-tauri/src/lib/git_tools.rs
```diff
@@ -1468,25 +1483,33 @@ fn fetch_single_commit_gix(
...
+                .map_err(|e| format!("Failed to prepare fetch: {}", e))?
+                .with_shallow(shallow.clone())
+                .receive(gix::progress::Discard, should_interrupt)
+                .map_err(|e| format!("Failed to receive: {}", e))
+        },
+        3,
+        std::time::Duration::from_millis(500),
+    ).map_err(|e| {
+        format!(
+            "fetch_single_commit_gix failed after 3 attempts: {}",
```
alirana01, line 1509: very small oversight but can make our lives easier later to actually log the attempts variable here

alirana01 (approved):
LGTM! Thanks just minor improvements we can maybe even do it in another PR as this one seems to be functionally fine

src-tauri/src/lib/git_tools.rs
Hahihula: this part of code is executed only if all the attempts fails.

PR #711, opened 2026-04-17
Eim 720: remove scoop from prerequisities installation

Hahihula:
manually tested on windows and it's reliably working, both cli and gui.
and offline installer:
image (11)

### DIFF
--- a/src-tauri/src/gui/commands/prequisites.rs
+++ b/src-tauri/src/gui/commands/prequisites.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use crate::gui::{app_state::get_settings_non_blocking, ui::send_message};
 use log::{error, warn};
 use rust_i18n::t;
@@ -111,8 +113,22 @@ pub fn install_prerequisites(app_handle: AppHandle) -> bool {
 
     // Spawn tokio task to avoid blocking the UI
     let app_handle_clone = app_handle.clone();
-    task::spawn_blocking(move || {
-        match idf_im_lib::system_dependencies::install_prerequisites(unsatisfied_prerequisites) {
+    let settings = match get_settings_non_blocking(&app_handle) {
+      Ok(settings) => settings,
+      Err(err) => {
+          let error_msg = t!("gui.system_dependencies.error_getting_settings", error = err.to_string()).to_string();
+          send_message(
+              &app_handle,
+              error_msg.clone(),
+              "error".to_string(),
+          );
+          error!("{}", error_msg);
+          return false;
+      }
+    };
+    let tool_install_directory = PathBuf::from(settings.tool_install_folder_name.clone().expect("Tools install folder not defined"));
+    tokio::spawn(async move {
+        match idf_im_lib::system_dependencies::install_prerequisites(unsatisfied_prerequisites, tool_install_directory).await {
             Ok(_) => {
                 // Emit event to notify frontend that installation is complete
                 let _ = app_handle_clone.emit("prerequisites-install-complete", json!({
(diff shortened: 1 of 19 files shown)


src-tauri/src/gui/commands/prequisites.rs
```diff
@@ -111,8 +113,22 @@ pub fn install_prerequisites(app_handle: AppHandle) -> bool {
...
+          send_message(
+              &app_handle,
+              error_msg.clone(),
+              "error".to_string(),
+          );
+          error!("{}", error_msg);
+          return false;
+      }
+    };
+    let tool_install_directory = PathBuf::from(settings.tool_install_folder_name.clone().expect("Tools install folder not defined"));
```
alirana01, line 129: I think gemini comment partially makes sense to handle the error gracefully for the gui

src-tauri/src/gui/commands/prequisites.rs
```diff
@@ -180,33 +196,49 @@ pub fn python_sanity_check(app_handle: AppHandle, python: Option<&str>) -> Vec<C
...
                 error_msg.clone(),
                 "error".to_string(),
             );
             error!("{}", error_msg);
-            false
+            return false;
         }
-    }
+    };
+    let tool_install_directory = PathBuf::from(settings.tool_install_folder_name.clone().expect("Tools install folder not defined"));
```
alirana01, line 215: similarly here maybe we can handle this case properly like handling the `None` case gracefully or some `false` for error cases with proper message.

src-tauri/src/lib/offline_installer.rs
```diff
@@ -376,80 +24,81 @@ fn setup_scoop_packages(scoop_path: &Path, scoop_command: &Path) -> Result<(), S
...
-        "linux" | "macos" => {
-            info!("On POSIX system, we do not provide prerequisites. Please ensure you have the necessary tools installed.");
+            // Python is expected to be in archive_dir as cpython-X.Y.Z+-x86_64-pc-windows-msvc-install_only.tar.gz
+            let python_archive = archive_dir.path().join("cpython-3.12.8+20250107-x86_64-pc-windows-msvc-install_only.tar.gz");
+            if !python_archive.exists() {
+                return Err(format!("Python archive not found in archive: {}", python_archive.display()));
+            }
+
+            // Copy python archive to tools_dir for installation
+            let python_in_tools = tools_dir.join("cpython-3.12.8+20250107-x86_64-pc-windows-msvc-install_only.tar.gz");
```
alirana01: This seems to be different than the one we have in `system_dependencies.rs` is it intentional?

src-tauri/src/offline_installer_builder.rs
```diff
@@ -742,23 +743,28 @@ async fn main() {
...
+                    error!("Failed to get latest Git for Windows URL: {}", err);
+                    return;
+                }
+            };
+
+            let prereq_list = vec![
+                // Git - portable Windows distribution (fetched dynamically from GitHub)
+                (git_url, git_filename),
+                // Python - standalone Windows distribution (matches system_dependencies.rs)
+                ("https://github.com/astral-sh/python-build-standalone/releases/download/20260414/cpython-3.11.15+20260414-x86_64-pc-windows-msvc-install_only.tar.gz".to_string(), "cpython-3.11.15+20260414-x86_64-pc-windows-msvc-install_only.tar.gz".to_string()),
```
alirana01: again if this is the same URL as we have in `system_dependencies.rs` this should become a module level constant

src-tauri/src/lib/offline_installer.rs
```diff
@@ -376,80 +24,81 @@ fn setup_scoop_packages(scoop_path: &Path, scoop_command: &Path) -> Result<(), S
...
+            // Copy git archive to tools_dir for installation
+            std::fs::copy(&git_archive_path, tools_dir.join(git_archive_path.file_name().unwrap()))
+                .map_err(|e| format!("Failed to copy git archive: {}", e))?;
+
+            // Install git from the copied archive
+            let rt = tokio::runtime::Runtime::new()
+                .map_err(|e| format!("Failed to create runtime: {}", e))?;
+            match rt.block_on(crate::system_dependencies::install_git_from_downloaded(
+                tools_dir.clone(),
+                None, // Archive will be found in tools_dir by install_git_from_downloaded
```
alirana01: This function is called from `run_wizzard_run` (async) and `start_offline_installation` (async) but creates its own Tokio runtime and blocks. This should theoretically hit some deadlocks or cause panic
Can we make `install_prerequisites_offline` an `async fn` so it can `.await` directly, matching the pattern already established for `install_prerequisites`.

src-tauri/src/lib/system_dependencies.rs
```diff
@@ -641,140 +640,530 @@ pub fn check_qemu_prerequisites() -> Result<Vec<&'static str>, String> {
...
-                    debug!("Could not get scoop path");
-                    return Err(String::from("Could not get scoop path"));
-                }
+            let (git_url, git_filename) = get_latest_git_for_windows_url().await?;
+            debug!("Downloading Git from {}", git_url);
+
+            // Download git portable archive
+            download_file_and_rename(
+                &git_url,
+                tools_dir.to_str().unwrap(),
```
alirana01: `to_str()` returns `None` for non-UTF-8 paths (valid on Windows). These `unwrap()` calls will panic on paths containing non-UTF-8 characters. I think `.to_string_lossy()` is a better choice

src-tauri/src/lib/system_dependencies.rs
```diff
@@ -641,140 +640,530 @@ pub fn check_qemu_prerequisites() -> Result<Vec<&'static str>, String> {
...
-                None => {
-                    debug!("Could not get scoop git path");
-                    "".to_string()
-                }
+
+            debug!("Extracting Git archive at {}", git_archive.display());
+
+            // Extract the archive
+            decompress_archive(
+                git_archive.to_str().unwrap(),
```
alirana01: Same here I think `.to_string_lossy()` is a better choice

src-tauri/src/lib/system_dependencies.rs
```diff
@@ -641,140 +640,530 @@ pub fn check_qemu_prerequisites() -> Result<Vec<&'static str>, String> {
...
+            const PYTHON_URL: &str = "https://github.com/astral-sh/python-build-standalone/releases/download/20260414/cpython-3.11.15+20260414-x86_64-pc-windows-msvc-install_only.tar.gz";
+
+            const PYTHON_FILENAME: &str = "cpython-3.11.15+20260414-x86_64-pc-windows-msvc-install_only.tar.gz";
+
+            debug!("Downloading Python from {}", PYTHON_URL);
+
+            // Download python archive
+            download_file_and_rename(
+                PYTHON_URL,
+                tools_dir.to_str().unwrap(),
```
alirana01: Same here I think `.to_string_lossy()` is a better choice

src-tauri/src/lib/system_dependencies.rs
```diff
@@ -641,140 +640,530 @@ pub fn check_qemu_prerequisites() -> Result<Vec<&'static str>, String> {
...
+            const PYTHON_FILENAME: &str = "cpython-3.11.15+20260414-x86_64-pc-windows-msvc-install_only.tar.gz";
+
+            let python_archive = python_archive_path.unwrap_or_else(|| tools_dir.join(PYTHON_FILENAME));
+
+            debug!("Extracting Python to {}", tools_dir.display());
+
+            // Extract the archive
+            decompress_archive(
+                python_archive.to_str().unwrap(),
+                tools_dir.to_str().unwrap(),
```
alirana01: Same here I think `.to_string_lossy()` is a better choice

alirana01 (commented):
Only some minor changes I think and some clarifications for my understanding

src-tauri/src/gui/commands/prequisites.rs
Hahihula: It does not. If you look and broader flow the settings.tool_install_folder_name can not be empty and if it is we did something terribly wrong and it SHOULD explode.

src-tauri/src/lib/offline_installer.rs
Hahihula: Thanks! Good catch.

src-tauri/src/lib/offline_installer.rs
Hahihula: you are right. That's the correct approach. It accidentally works because it's on the windows part where it runs in separated thread.

alirana01 (approved):
LGTM!

PR #719, opened 2026-04-21
EIM-661: Read hardcoded string directly from vue i18n files for tests

alirana01:
## Summary

GUI tests no longer hardcode UI copy. They now resolve strings through a thin test-only helper (`tests/helpers/i18n.js`) that reads the same `src/locales/{en,cn}.json` files the production app loads via `vue-i18n`. Change a string in the locale file and the tests follow automatically.

## Why a helper (and not `vue-i18n` directly)

- **No prod helper to reuse.** The app just renders translations via `vue-i18n`'s `t()` / `$t()`; nothing in `src/` resolves keys for comparison. That's a test-only need.
- **`vue-i18n` could do the lookup, but lookup isn't the hard part.** The helper's value is the two wrappers that sit *on top of* lookup:
  - `matchable(key)` — returns the portion before the first `{var}` or `<tag>`. Needed because Selenium's `getText()` strips HTML and dynamic values are unknown at assert time.
  - `xpathText(key)` — returns the longest apostrophe-free segment. Needed because `GUITestRunner` builds XPath 1.0 expressions with single-quoted literals, which can't contain `'` (e.g. "Don't show…").
  Neither is something `vue-i18n` offers; they'd live in a test wrapper regardless.
- **Fail-loud is a feature in tests.** Prod has `fallbackLocale: "en"` and silently returns the key on miss. Tests *want* a missing/renamed key to throw immediately thats what catches locale drift. The helper throws on missing keys and missing `{var}` values.
- **Node 20.x CI compatibility.** Uses `fs.readFileSync` + `JSON.parse` instead of `import … with { type: "json" }`, which isn't stable on every Node 20.x minor we run.

Net: the lookup slice of the helper is ~20 lines of boring code; the Selenium-specific bits are the actual payoff.

## What changed

- **New:** `tests/helpers/i18n.js`exports `tGui`, `matchable`, `xpathText`.
- **Refactored (8 files):** all `GUI*.test.js` scripts under `tests/scripts/` to resolve UI assertions through the helper instead of inline string literals.
- **No production code changes** beyond flagging a couple of missing/hardcoded strings in components for follow-up (noted inline as `NOTE(EIM-661)`).

## Scope notes

- GUI/Vue side only. Rust/CLI strings are not touched yet

### DIFF
--- a/tests/helpers/i18n.js
+++ b/tests/helpers/i18n.js
@@ -1,57 +1,64 @@
 /**
  * i18n helper for GUI tests.
  *
- * Reads the same vue-i18n JSON locale files the production GUI uses
- * (src/locales/{en,cn}.json) so tests assert against a single source of
- * truth: change a string in the locale file and the test follows
- * automatically.
+ * Resolves UI strings through the same `vue-i18n` library the production
+ * GUI uses, fed from the same locale files (`src/locales/{en,cn}.json`),
+ * so tests assert against a single source of truth: change a string in
+ * the locale file and the tests follow automatically.
  *
  * Exports:
- *   - tGui(path, vars?, locale?) — resolve a dotted i18n path to its
- *     rendered string, substituting {var} placeholders from `vars`.
- *     Throws if the path is missing, if the path does not resolve to a
- *     string, or if any {var} placeholder has no matching entry in
- *     `vars`.
- *   - matchable(path, locale?) — return the portion of the translated
- *     string that precedes the first {var} placeholder, trimmed. Useful
- *     for substring assertions (findByText / .include(...)) when the
- *     dynamic value is unknown or noisy.
+ *   - tGui(path, vars?, locale?) — translate a dotted i18n path via
+ *     `vue-i18n`'s `t()`. Throws if the key is missing (i.e. `t()`
+ *     returned the path verbatim).
+ *   - matchable(path, locale?) — returns the portion of the *raw*
+ *     locale value that precedes the first `{var}` placeholder or
+ *     `<tag>`, trimmed. Used for substring assertions when Selenium's
+ *     `getText()` strips inline HTML or when the dynamic value is
+ *     unknown at assert time.
+ *   - xpathText(path, locale?) — returns the longest apostrophe-free
+ *     segment of the raw locale value. `GUITestRunner` builds XPath 1.0
+ *     `contains(text(), '...')` expressions, whose single-quoted literal
+ *     cannot embed a real `'`.
  *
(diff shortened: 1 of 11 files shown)


tests/helpers/i18n.js
Hahihula, line 1: i have a little issue with this.
seems like reimplementing the wheel. on the UI we use vue-i18n library, and it would seems reasonable to use the same one and not reimplement the logic again.
most of this file could be replaced by simple:
```
import { createI18n } from "vue-i18n";
import en from "../../src/locales/en.json" assert { type: "json" };
import cn from "../../src/locales/cn.json" assert { type: "json" };

const i18n = createI18n({ locale: "en", messages: { en, cn } });
export const tGui = (key, vars, locale = "en") =>
  i18n.global.t(key, vars, { locale });
```
and it will not need to be maintained separately and it will cover all the posibilities of the vue-i18n not just selected subset as you do now.

As main reason for this PR was to make test less brittle, this introduction of new logic does not seems right to me.

Hahihula (changes requested):

tests/helpers/i18n.js
alirana01: `tGui` is now a thin wrapper around `i18n.global.t()` from a single `createI18n()` instance fed the same locale JSON the app loads. The hand-rolled `walk()` and `{var}` regex are gone. Added `vue` + `vue-i18n` to `tests/package.json`, pinned to the app's versions. Fail-loud preserved via `fallbackLocale: false` + an `out === key` check.
`matchable()` and `xpathText()`. They aren't lookup logic — they're Selenium adapters. `getText()` strips inline HTML so we can't match against a full `t()` output for keys like `<strong>{name}</strong>`, and `GUITestRunner.findByText` builds XPath 1.0 `contains(text(), '...')` expressions whose single-quoted literal can't embed apostrophes (e.g. "Don't show…"). Both still source from the locale file via the new vue-i18n setup, just walking the raw template before interpolation.

Hahihula (dismissed):
Thanks! LGTM!

tests/node_modules.zip
Hahihula, line 1: Can you please explain the inclusion of this zip?

alirana01: I acutally didn't include this in this PR it is a pre-existing zip as windows side things skip npm ci and extract this archive since we added two new dependencies we needed to update the archive. I think initially Fabricio did it to avoid any conflicts and slow downs in the tests on runners

Hahihula (approved):
LGTM! Thanks for the change.

PR #742, opened 2026-04-29
EIM-750 Split offline installer pipeline

Hahihula:
This should fix the offline installer build on arm runners. But seems to be working only with the offline_installer_builder from the branch https://github.com/espressif/idf-im-ui/pull/711
(we need both changes)
image

The disadvantage of this approach is that the build is much slower to build all the versions of offline installer.
Also this should improve the visibility by producing "dashboard" like output
image

Single version build:
https://github.com/espressif/idf-im-ui/actions/runs/25100079847
All versions build: (runs almost 2 hours)
https://github.com/espressif/idf-im-ui/actions/runs/25102633559
image

### DIFF
--- a/.github/workflows/build_offline_installer_archives.yaml
+++ b/.github/workflows/build_offline_installer_archives.yaml
@@ -12,12 +12,12 @@ on:
       idf_version:
         required: false
         type: string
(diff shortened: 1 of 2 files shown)


alirana01 (approved):
Well LGTM! but it is ci so good luck

Issue #759, opened 2026-05-04
linuxdeploy AppRun.wrapped exports PYTHONHOME/PYTHONPATH/...* for the eim process; vars leak to subprocesses and break the Python sanity check (pip, venv, ctypes, SSL/HTTPS) (EIM-766)


FireDeveloper:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

Detect pip and other python prerequisites.

### Actual behavior (suspected bug)

It detects only python and nothing under it.

### Error logs or terminal output

```shell

```

### Steps to reproduce the behavior

  1. On a Linux host with a working Python 3 / pip, launch the eim GUI.                                                                                                                                                                                                                                                     
  2. Click into the install wizard and reach the Python sanity check step.
  3. All six rows (Python Version / pip / venv / Standard Library / ctypes / SSL/HTTPS) FAIL except the first. The detailed message recommends python3 -m ensurepip --upgrade, which is unrelated to the actual cause.                                                                                                      
                                                                                                                                                                                                                                                                                                                            
  Reproducing the underlying error in a terminal:                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
  $ python3 -m pip --version                                                                                                                                                                                                                                                                                                
  pip 26.1 from /usr/lib/python3.14/site-packages/pip (python 3.14)
                                                                                                                                                                                                                                                                                                                            
  $ env PYTHONHOME=/path/to/AppDir/usr/ python3 -m pip --version                                                                                                                                                                                                                                                            
  Fatal Python error: Failed to import encodings module                                                                                                                                                                                                                                                                     
  Python runtime state: core initialized                                                                                                                                                                                                                                                                                    
  ModuleNotFoundError: No module named 'encodings'                
                                                                                                                                                                                                                                                                                                                            
  Same output as the wizard's [FAIL] rows, exactly.

### Project release version

v0.12.0

### System architecture

Intel/AMD 64-bit (modern PC, older Mac)

### Operating system

Linux

### Operating system version

CachyOS

### Shell

ZSH

### Additional context

Claude figured out this too.  It's the last one, after that it installed idf successfully, and build a project.

[...]

Issue #757, opened 2026-05-04
AppImage bundles libnghttp2.so.14; subprocesses inherit LD_LIBRARY_PATH and abort on system libcurl symbol mismatch — silently breaks the prerequisite check on Arch / CachyOS (EIM-764)


FireDeveloper:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

Resolve prerequisites correctly

### Actual behavior (suspected bug)

Reporting that libusb and libffi are missing.

### Error logs or terminal output

```shell

```

### Steps to reproduce the behavior

1. On a current Arch / CachyOS host (same libnghttp2/libcurl mismatch direction), launch the eim GUI from the AppImage.
*Apply fixes:
https://github.com/espressif/idf-im-ui/issues/755
https://github.com/espressif/idf-im-ui/issues/756
2. Click into the install wizard so the prerequisite check runs.
3. The wizard reports libusb and libffi as missing, regardless of whether they are actually installed:                                                                                                                                                                                                                    
  $ pacman -Q libusb libffi
  libusb 1.0.29-1.1
  libffi 3.5.2-1.1
4. From a terminal, with the AppImage's LD_LIBRARY_PATH simulated, the underlying failure is visible:                                                                                                                                                                                                                     
  $ env LD_LIBRARY_PATH=/path/to/AppDir/usr/lib pacman -Qs | grep libusb
  pacman: symbol lookup error: /usr/lib/libcurl.so.4: undefined symbol:
    nghttp2_option_set_no_rfc9113_leading_and_trailing_ws_validation
4. Without the bundled libnghttp2 shadowing system libnghttp2, the same query succeeds:                                                                                                                                                                                                                                   
  $ pacman -Qs | grep libusb                                                                                                                                                                                                                                                                                                
  local/libusb 1.0.29-1.1                                         
  local/libusb-compat 0.1.8-2.1                                                                                                                                                                                                                                                                                             
  local/libusbmuxd 2.1.1-2.1

### Project release version

v0.12.0

### System architecture

Intel/AMD 64-bit (modern PC, older Mac)

### Operating system

Linux

### Operating system version

CachyOS

### Shell

ZSH

### Additional context

Report by claude:

### Summary                                                         
[...]

Issue #756, opened 2026-05-04
GUI splash spinner never advances on non-NVIDIA Linux GPUs (vmwgfx, possibly others) (EIM-763)


FireDeveloper:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

After the splash, the wizard's first page (version selection / prerequisites / mirrors) renders.

### Actual behavior (suspected bug)

Stuck on splash screen with frozen spinner.

### Error logs or terminal output

```shell

```

### Steps to reproduce the behavior

1. Apply the libwayland workaround (other issue) so the EGL init succeeds.
https://github.com/espressif/idf-im-ui/issues/755
2. Launch eim on a host using the vmwgfx Mesa driver (e.g. a VMware Workstation / ESXi VM, or in a CI runner with VMware GPU).
3. Window opens, splash logo paints, spinner appears and never stops.                                                                                                                                                                                                                                                     
4. The Rust backend completes its work — internal log shows mirror probes finishing, prerequisite checks passing — but the WebView never advances to the wizard.

### Project release version

v0.12.0

### System architecture

Intel/AMD 64-bit (modern PC, older Mac)

### Operating system

Linux

### Operating system version

CachyOS

### Shell

ZSH

### Additional context

Claude figured this too. Now UI is fully functional:

### Summary                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
The eim GUI binary already ships a workaround that disables WebKitGTK's DMA-BUF renderer when NVIDIA is detected:                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                            
  Nvidia detected: disabled WebKitGTK DMA-BUF renderer (WEBKIT_DISABLE_DMABUF_RENDERER=1)                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  (visible in strings eim, gated on the presence of /proc/driver/nvidia/version).                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
[...]

Issue #755, opened 2026-05-04
GUI AppImage shows blank window on Linux (CachyOS) (EIM-762)


FireDeveloper:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

GUI renders the eim wizard. 

### Actual behavior (suspected bug)

Window is blank (white)

### Error logs or terminal output

```shell
stderr shows:
Running GUI...
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

### Steps to reproduce the behavior

1. Download eim-gui-linux-x64.AppImage from the latest release.                                                                                                                                                                                                                                                           
2. Install it via shelly and run it.
3. Window appears, stays blank, process is in a half-alive state.                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                            
Run with the AppImage's stderr visible (e.g. from a terminal) to see the EGL_BAD_PARAMETER line. 

### Project release version

v0.12.0

### System architecture

Intel/AMD 64-bit (modern PC, older Mac)

### Operating system

Linux

### Operating system version

CachyOS (more bellow)

### Shell

ZSH

### Additional context

Claude figured it out. Bellow you can find it's words, it may make sense to you...

### Summary                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  eim-gui-linux-x64.AppImage opens a blank window and aborts internally on a current Arch-family Linux. The cause is that the AppImage bundles its own libwayland-client.so.0, libwayland-cursor.so.0, libwayland-egl.so.1, libwayland-server.so.0 and prepends them via LD_LIBRARY_PATH (set by the linuxdeploy            
  AppRun.wrapped). The host's libEGL (Mesa) then loads those older bundled wayland symbols and fails:
                                                                                                                                                                                                                                                                                                                            
  Running GUI...                                                  
  Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
                                                                                                                                                                                                                                                                                                                            
[...]

PR #761, opened 2026-05-05
Eim-762: add pacman support

Hahihula:
should solve #755 #756 #757 #759
@achxkloel tested it on Arch

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -644,6 +644,157 @@ jobs:
           cp "$RPM_FILE" "eim-cli-${{ matrix.package_name }}.rpm"
           gh release upload "${{ github.event.release.tag_name }}" "eim-cli-${{ matrix.package_name }}.rpm" --repo "${{ github.repository }}" --clobber
 
+  # ============================================================
+  # Build CLI pacman packages inside Arch container
+  # ============================================================
+  build-cli-pacman:
+    name: Build CLI pacman (${{ matrix.package_name }})
+    needs: [build-cli-linux, build-cli-linux-armv7]
+    if: needs.build-cli-linux.result == 'success'
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write
+    container:
+      image: archlinux:latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            package_name: linux-x64
+            target: "x86_64-unknown-linux-musl"
+            pacman_arch: x86_64
+          - os: ubuntu-24.04-arm
+            package_name: linux-aarch64
+            target: "aarch64-unknown-linux-musl"
+            pacman_arch: aarch64
+          - os: ubuntu-latest
+            package_name: linux-armv7
+            target: "armv7-unknown-linux-musleabihf"
+            pacman_arch: armv7
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install Arch dependencies
+        run: |
+          pacman -Sy --noconfirm base-devel curl gh git
+
+      - name: Install Rust toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          source "$HOME/.cargo/env"
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+
+      - name: Download built binary artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: rpm-build-input-${{ matrix.package_name }}-${{ (github.event_name == 'release' && github.ref_name) || github.run_number }}
+          path: pacman-input/
+
+      - name: Prepare build tree
+        shell: bash
+        run: |
+          set -euo pipefail
+          source "$HOME/.cargo/env"
+
+          TARGET="${{ matrix.target }}"
+          RELEASE_DIR="src-tauri/target/${TARGET}/release"
+          mkdir -p "$RELEASE_DIR"
+
+          # Copy binary from the downloaded artifact into the expected location
+          cp pacman-input/release_cli/${{ matrix.package_name }}/eim "$RELEASE_DIR/eim"
+          chmod +x "$RELEASE_DIR/eim"
+
+          # Copy completions if present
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -644,6 +644,157 @@ jobs:
...
+          builddate = $(date +%s)
+          packager = CI
+          arch = $PACMAN_ARCH
+          license = MIT
+          depend = git
+          depend = wget
+          depend = libusb
+          depend = libffi
+          depend = openssl
+          EOF
```
alirana01: Is it working fine maybe the pacman doesn't mind the extra leading spaces that are created by this in .PKGINFO but i can recommend verifying this loaclly.

.github/workflows/build.yaml
```diff
@@ -644,6 +644,157 @@ jobs:
...
+
+                    # Create .INSTALL (optional post-install script)
+                    cat > "$PKG_DIR/.INSTALL" << 'EOF'
+          post_install() {
+            echo "Installing eim completion files..."
+          }
+          EOF
```
alirana01: similarly here not sure if the weird indentation in created files will work or not

.github/workflows/build.yaml
```diff
@@ -644,6 +644,157 @@ jobs:
...
+
+      - name: Install Arch dependencies
+        run: |
+          pacman -Sy --noconfirm base-devel curl gh git
+
+      - name: Install Rust toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          source "$HOME/.cargo/env"
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
```
alirana01: Do we need to setup the toolchain are we building it again? Also same in gui side. I thought we are only packaging pre-built binaries.
and if this is related to change at line 699 `source "$HOME/.cargo/env"` i guess that can also be removed

.github/workflows/build.yaml
```diff
@@ -644,6 +644,157 @@ jobs:
...
+          EOF
+
+                    # Create .INSTALL (optional post-install script)
+                    cat > "$PKG_DIR/.INSTALL" << 'EOF'
+          post_install() {
+            echo "Installing eim completion files..."
+          }
+          EOF
+
+          # Create the package
```
alirana01: Since this is already running on archlinux container why not use `makepkg`. for reference check [this](https://wiki.archlinux.org/title/Makepkg) 
It should produce correct package and reduce the work in this file as well.

.github/workflows/update-linux-repos.yml
```diff
@@ -284,3 +284,140 @@ jobs:
...
+          }
+
+          generate_repo "x86_64" "x86_64"
+          generate_repo "aarch64" "aarch64"
+          generate_repo "armv7" "armv7"
+
+      - name: Upload repository to S3
+        run: |
+          # Upload the entire repository structure
+          aws s3 sync pacman-repo/ s3://espdldata/dl/eim/pacman/ --acl public-read --delete
```
alirana01: this `--delete` is tricky i guess it will wipe everything in case something goes wrong I dont think we use it for apt or rpm at teh moment so i would avoid that unless very necessary

.github/workflows/update-linux-repos.yml
```diff
@@ -284,3 +284,140 @@ jobs:
...
+            ls -la pacman-repo/x86_64/ || echo "No x86_64 packages"
+            echo "aarch64 packages:"
+            ls -la pacman-repo/aarch64/ || echo "No aarch64 packages"
+            echo "armv7 packages:"
+            ls -la pacman-repo/armv7/ || echo "No armv7 packages"
+          else
+            echo "No pacman artifacts downloaded"
+          fi
+
+      - name: Generate pacman repository database
```
alirana01: This step doesn't create a valid pacman sync database. Extracting packages with `tar` into a `db/` folder is not how pacman repos work. Pacman expects a database file (e.g. `espressif.db.tar.gz`) generated by [`repo-add`](https://man.archlinux.org/man/repo-add.8):

```bash
repo-add espressif.db.tar.gz *.pkg.tar.gz
```

Without this, `pacman -Sy` will look for `espressif.db` at the Server URL, get a 404, and fail.
Since this job runs on Ubuntu, you'll need to either run this step in an `archlinux:latest` container or install pacman tools on Ubuntu.

.github/workflows/update-linux-repos.yml
```diff
@@ -284,3 +284,140 @@ jobs:
...
+          # Function to generate repo for an architecture
+          generate_repo() {
+            local arch_dir="$1"
+            local arch_name="$2"
+            if [ -d "$arch_dir" ] && [ "$(ls -A $arch_dir/*.pkg.tar.gz 2>/dev/null)" ]; then
+              echo "Generating database for $arch_name..."
+              # Create the repo database using tar (standard tarball for pacman)
+              # Pacman uses a simple tarball with a database directory structure
+              mkdir -p "$arch_dir/db"
+              cd "$arch_dir"
```
alirana01: `cd "$arch_dir"` but then no `cd ..` back, so the 2nd and 3rd `generate_repo` calls may look in the wrong directory.

docs/src/general_info.md
```diff
@@ -222,9 +223,47 @@ sudo dnf upgrade eim
...
+
+# Or install GUI (includes CLI)
+sudo pacman -S eim-gui
+```
+
+> **Note:** The GUI package depends on `gtk4`, `libadwaita`, and `webkit2gtk-4.1`, which may require additional setup on some distributions.
+
+To update EIM later:
+```bash
+sudo pacman -Syu eim
```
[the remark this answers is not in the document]
alirana01: so partially correct as the package eim doesn't exist `eim-cli` and `eim-gui` i guess are the correct ones

alirana01 (changes requested):
I think there are some issues here i was trying to read the docs for pacman on arch and some comments left only on cli step but may also apply to gui build for pacman.

.github/workflows/build.yaml
Hahihula: it requires a non-root user and a PKGBUILD

alirana01 (approved):
Good luck 🤞

PR #798, opened 2026-05-19
abandoning actions/download-artifact because they are so stupid that they have hardcoded limit to 100 artifacts so you will never get the ones on the end of the alphabet

Hahihula:
abandoning actions/download-artifact because they are so stupid that they have hardcoded limit to 100 artifacts so you will never get the ones on the end of the alphabet


alirana01 (dismissed):
approved with extreme annoyance towards you know who

alirana01 (approved):

Issue #796, opened 2026-05-19
Activation scripts set ESP_IDF_VERSION without patch component, breaking idf.py on v6.0.1 (EIM-796)


Babyeggz:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

After sourcing `~/.espressif/tools/activate_idf_v6.0.1.fish` (or the `.sh` equivalent), `ESP_IDF_VERSION` should be set to `6.0.1` (the full MAJOR.MINOR.PATCH version), and `idf.py --version` should run successfully and print `ESP-IDF v6.0.1`.

### Actual behavior (suspected bug)

The activation scripts generated by EIM export `ESP_IDF_VERSION` without the patch component:

- **bash** (`activate_idf_v6.0.1.sh`): sets `ESP_IDF_VERSION=6.0` (MAJOR.MINOR only)
- **fish** (`activate_idf_v6.0.1.fish`): sets `ESP_IDF_VERSION=""` (empty string)

On v6.0.1, this breaks `idf.py` outright, because `idf_component_manager/idf_extensions.py` line 385 does:

​```python
if Version.coerce(os.getenv('ESP_IDF_VERSION')) >= Version.coerce('6.0.1'):
​```

`Version.coerce('6.0')` and `Version.coerce('')` both raise `ValueError: Version string lacks a numerical component`, so `idf.py` cannot run at all — not even `idf.py --version`.

**Root cause:** `$IDF_PATH/tools/cmake/version.cmake` contains all three components:

​```cmake
set(IDF_VERSION_MAJOR 6)
set(IDF_VERSION_MINOR 0)
set(IDF_VERSION_PATCH 1)
​```

But the activation scripts only parse MAJOR and MINOR. In the fish script's `parse_cmake_version` function, the if/else-if chain matches only `IDF_VERSION_MAJOR` and `IDF_VERSION_MINOR`, never `IDF_VERSION_PATCH`. The bash script has the same omission, which produces `6.0` instead of `6.0.1`.

**Suggested fix:** parse all three components from `version.cmake` and combine them as `$major.$minor.$patch` in both `.sh` and `.fish` activation script templates.

[...]

PR #834, opened 2026-05-27
EIM-796: Update IDF version to include patch version

Hahihula:
should solve #796

### DIFF
--- a/src-tauri/bash_scripts/activate_idf_template.fish
+++ b/src-tauri/bash_scripts/activate_idf_template.fish
@@ -15,6 +15,7 @@ function parse_cmake_version
 
     set major ""
     set minor ""
+    set patch ""
 
     # Read the file and extract version numbers
     while read -l line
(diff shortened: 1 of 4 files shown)


src-tauri/bash_scripts/activate_idf_template.fish
```diff
@@ -15,6 +15,7 @@ function parse_cmake_version
 
     set major ""
     set minor ""
+    set patch ""
 
     # Read the file and extract version numbers
     while read -l line
```
XDanielPaul, line 21: Nit:

Not really related to the PR, but if all version components are set, there's no need to finish reading the file

```bash
if test -n "$major"; and test -n "$minor"; and test -n "$patch"
   break
end
```

src-tauri/bash_scripts/activate_idf_template.fish
XDanielPaul, line 39: Nit:

Suggesting `; or` instead of `-o`

`if test -z "$major"; or test -z "$minor"`

XDanielPaul (approved):
LGTM

PR #835, opened 2026-05-28
EIM-827 fix openocd script path

Hahihula:


src-tauri/src/lib/idf_tools.rs
```diff
@@ -699,6 +699,48 @@ pub fn get_tools_export_paths_from_list(
...
+  installed_tools: HashMap<String, (String, Download)>,
+  tools_install_path: &str,
+) -> Vec<(String, String)> {
+    let mut vars: Vec<(String, String)> = Vec::new();
+
+    for (tool_name, (_version, _download)) in installed_tools {
+        if let Some(tool) = _tools_file.tools.iter().find(|t| t.name == tool_name) {
+            for (var_name, var_value) in &tool.export_vars {
+                // Replace ${TOOL_PATH} with the actual tools_install_path
+                let processed_value = var_value.replace("${TOOL_PATH}", tools_install_path);
```
kolipakakondal: Quick check on my understanding here, I might be missing something...

It looks like ${TOOL_PATH} is being replaced with tools_install_path (e.g ~/.espressif/tools), but per the tools.json I think ${TOOL_PATH} is meant to be the per-tool, per-version install dir, i.e. <tools_install_path>/<tool_name>/<version>. So for OpenOCD it should expand to something like:

~/.espressif/tools/openocd-esp32/v0.12.0-esp32-20260424/openocd-esp32/share/openocd/scripts
rather than just ~/.espressif/tools/openocd-esp32/share/openocd/scripts. please check

kolipakakondal (dismissed):
LGTM, please do check the comment.

@brianignacio5 please do verify with the builds.

src-tauri/src/lib/idf_tools.rs
Hahihula: Thanks for the catch, should be solved now.

kolipakakondal (approved):
Thanks for the fix, LGTM.

brianignacio5:
> LGTM, please do check the comment.
> 
> @brianignacio5 please do verify with the builds.

LGTM to me too. Tested on MacOS M1 with CLI

PR #848, opened 2026-06-03
EIM-837: fix execution of activation script

Hahihula:
image (19)


brianignacio5:
Tested on MacOS and LGTM

PR #859, opened 2026-06-04
EIM-761: workaround for small /tmp partition added to docs

Hahihula:


docs/src/faq.md
```diff
@@ -191,6 +191,41 @@ To use PowerShell, simply run the generated PowerShell profile script - it is cr
...
+Then run `eim` from the same terminal session. You can remove `~/tmp_build` after installation completes.
+
+To check how much space your `/tmp` has:
+
+```bash
+df -h /tmp
+```
+
+If it shows only 1–2 GB available, the workaround above is recommended, especially when installing large toolchains like `clang-esp32` or `riscv32`.
+
```
kolipakakondal, line 228: One suggestion: We could display the minimum required disk space for setting up EIM during the tools installation directory step. We already mention, "Ensure you have sufficient disk space" but explicitly stating the minimum required space would help users avoid issues like this.

kolipakakondal (approved):
LGTM

Issue #875, opened 2026-06-10
eim v0.13.1 fails python sanity check with a system-wide proxy configured (EIM-858)


zh5j:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

Sanity check should pass regardless of proxy config

### Actual behavior (suspected bug)

Fails python sanity check in the SSL/HTTPS step when HTTP(S)_PROXY and http(s)_proxy are set in `/etc/environment`, passes when no proxy set
(please see terminal output below)

### Error logs or terminal output

```shell
(with proxy)

2026-06-10 14:33:27 -  2 - 06 - INFO - Running IDF Installer Wizard...
2026-06-10 14:33:27 -  2 - 06 - INFO - ESP-IDF JSON file already exists at: "/home/abrobo/.espressif/tools/eim_idf.json"
2026-06-10 14:33:28 -  2 - 06 - INFO - All prerequisites are satisfied!
2026-06-10 14:33:28 -  2 - 06 - INFO - Running python sanity check
⠂ Doing something...                                                                                                                                                             2026-06-10 14:33:28 -  2 - 06 - INFO - Found python3
⠐ Doing something...                                                                                                                                                               [PASS] Python Version
  [PASS] pip
  [PASS] venv
  [PASS] Standard Library (json, os, platform)
  [PASS] ctypes
  [FAIL] SSL/HTTPS
         Hint: Check your network and proxy settings, or try to install OpenSSL development headers: sudo apt install libssl-dev (Debian/Ubuntu), then reinstall Python
Error executing CLI: Python sanity check failed. See the [FAIL] entries above for details.

(without proxy)

2026-06-10 14:40:54 -  2 - 06 - INFO - Running IDF Installer Wizard...
2026-06-10 14:40:54 -  2 - 06 - INFO - ESP-IDF JSON file already exists at: "/home/abrobo/.espressif/tools/eim_idf.json"
2026-06-10 14:40:55 -  2 - 06 - INFO - All prerequisites are satisfied!
2026-06-10 14:40:55 -  2 - 06 - INFO - Running python sanity check
⠂ Doing something...                                                                                                                                                             2026-06-10 14:40:55 -  2 - 06 - INFO - Found python3
[...]

PR #877, opened 2026-06-10
EIM-858: updated docs FAQ with info about systemwide proxy

Hahihula:
solves #875

### DIFF
--- a/docs/src/faq.md
+++ b/docs/src/faq.md
@@ -110,6 +110,38 @@ eim install --skip-prerequisites-check=true
 
 > ⚠️ **Warning:** Use this flag only if you are certain that all required prerequisites are correctly installed, as skipping the check may lead to installation failures later.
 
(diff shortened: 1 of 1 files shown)


zh5j:
Note for @Hahihula : the same issue comes up if the HTTP(S)_PROXY variables are set manually for the shell, and the NO_PROXY is missing; I'd suggest not naming this as "system proxy", rather, I think we can instruct the user to check their environment by `env|grep -i proxy` and add NO_PROXY if necessary. My working config:
```
http_proxy=http://127.0.0.1:8084/
https_proxy=http://127.0.0.1:8084/
HTTP_PROXY=http://127.0.0.1:8084/
HTTPS_PROXY=http://127.0.0.1:8084/
no_proxy=127.0.0.1,localhost
NO_PROXY=127.0.0.1,localhost
```

kolipakakondal (approved):

PR #899, opened 2026-06-19
EIM-739: fixed hardcodev idf versions in tests

Hahihula:


alirana01 (approved):
LGTM! just need to verify that tests are passing

PR #903, opened 2026-06-22
Eim 861 new eim idf json approach

Hahihula:
in this PR we are adding new approach to the version management which now takes care for unsucesfull or unfinished installations too.
on the version management scree nthe user clearly see status of the installation:
image
and actually if broken installation is detected, the user is warn on the installation start:
ESP-IDE Installation Manager
the same applies to the CLI:
note to see what the problems were,
Running
Running &#39;targetreleaseeim select
Running

## Summary by CodeRabbit

* **New Features**
  * Added installation status tracking for finished, in-progress, failed, repairing, and broken installations.
  * Added automatic detection of incomplete installations with options to repair, delete, dismiss, or skip them.
  * Added status labels to version lists, selection prompts, and management cards.
  * Added repair workflows for failed or broken installations.

* **Documentation**
  * Added guidance for handling incomplete installations and updated CLI and version-management documentation.

* **Tests**
  * Added CLI and GUI coverage for installation statuses and recovery workflows.

PR #905, opened 2026-06-23
EIM-878: telemetry update

Hahihula:
this is based on https://github.com/espressif/idf-im-ui/pull/903 branch so we track the new approach to the installation.


src-tauri/src/gui/commands/utils_commands.rs
```diff
@@ -293,30 +296,114 @@ pub async fn install_drivers() -> Result<(), String> {
...
+            let outcome = match outcome.as_deref() {
+                Some("success") => InstallOutcome::Success,
+                Some("failure") | None => InstallOutcome::Failure,
+                Some(other) => return Err(format!("Unknown outcome: {}", other)),
+            };
+            let kind = parse_error_kind(error_kind.as_deref())?;
+            let error = match (outcome, kind, error_message.as_deref()) {
+                (InstallOutcome::Failure, Some(k), Some(msg)) => Some(build_anyhow(k, msg)),
+                _ => None,
+            };
```
alirana01, line 373: **GUI failures drop error telemetry** is this working because if I look into the he Vue callers (`SimpleSetup`, `InstalationProgress`, `OfflineInstaller`) only send `error_message` on failure, so `kind` stays `None`, `error` is `None`, and `track_install_outcome` never populates `error_kind`, `error_message`, or `error_hash` even though the FAQs say we collect those.

CLI already does the right thing via `ErrorKind::from_message`

src-tauri/src/lib/telemetry.rs
```diff
@@ -16,53 +21,308 @@ static CLIENT: Lazy<Option<reqwest::Client>> = Lazy::new(|| {
...
+        extras: OutcomeExtras::default(),
+        subcommand: Some(subcommand.to_string()),
+    });
+}
+
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-struct TelemetryItem {
-    name: String,
+struct EventProps {
```
alirana01, line 246: Shouldn't we be sending the eim version nad system info with the telemetry?

src-tauri/src/lib/telemetry.rs
```diff
@@ -16,53 +21,308 @@ static CLIENT: Lazy<Option<reqwest::Client>> = Lazy::new(|| {
...
+    extras: OutcomeExtras,
+) {
+    if !is_enabled() {
+        return;
+    }
+    let (error_message, error_hash) = match (outcome, error) {
+        (InstallOutcome::Failure, Some(err)) => {
+            let raw = format!("{:#}", err);
+            let scrubbed = scrub_pii(&raw);
+            let hash = hash_short(&raw);
```
alirana01, line 201: we should be hashing the scrubbed string as mentioned in our faq just to avoid even making the hash contain the scrubbed fields to avoid any concerns from community

src-tauri/src/lib/telemetry.rs
```diff
@@ -16,53 +21,308 @@ static CLIENT: Lazy<Option<reqwest::Client>> = Lazy::new(|| {
...
+
+    let client = client.clone();
+    tokio::spawn(async move {
+        let _ = client
+            .post(&url)
+            .header("Content-Type", "application/x-json-stream")
+            .body(serial)
+            .send()
+            .await;
+    });
```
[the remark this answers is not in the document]
alirana01: @Hahihula this may make sense only if you think this can break teh installer or installation just because hte telemetry part had an exception

src/components/wizard_steps/InstalationProgress.vue
```diff
@@ -329,8 +329,10 @@ export default {
       this.timeStarted = new Date();
 
       try {
-        const eventName = this.is_fix_mode ? "GUI fix installation started" : "GUI wizard installation started";
-        await invoke("track_event_command", { name: eventName });
+        await invoke("track_event_command", {
+          event: "install_started",
+          mode: this.is_fix_mode ? "fix" : "wizard"
+        });
```
alirana01, line 335: Shouldn't we also send the version and installation IDs here as well?

alirana01 (commented):
There are some concerns on the PR please see if they are legit and can be fixed.

PR #802, opened 2026-05-21
fix(ui): align button theme colors (EIM-799)

HalfSweet:
## Description

This PR fixes the unexpected green/colored shadow around Naive UI primary buttons in the Vue frontend.

The issue was caused by globally overriding `.n-button` background color to Espressif red while Naive UI's internal primary theme variables still used the default green color. As a result, button border, focus, hover, and ripple states could render with mismatched green styling around red buttons.

Changes:
- Added Naive UI `themeOverrides` in `App.vue` to align primary button background, hover, pressed, focus, border, and ripple colors with the Espressif red theme.
- Removed the global `.n-button` background override from `main.css` to avoid conflicting with Naive UI component state styles.

Screenshot:

图片

## Related

N/A

## Testing

- Ran `npm run build` successfully.
- Verified the local frontend page at `http://127.0.0.1:1420/#/basic-installer`.
- Confirmed the affected primary button now uses red theme values for background, hover/focus border, and ripple color instead of the default green state color.

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [x] 🚨 This PR does not introduce breaking changes.
- [ ] All CI checks (GH Actions) pass.
- [x] Documentation is updated as needed.
- [x] Tests are updated or added as necessary.
- [x] Code is well-commented, especially in complex areas.
- [x] Git history is clean — commits are squashed to the minimum necessary.

PR #906, opened 2026-06-23
Fix/button theme shadow

Hahihula:
based on #802


alirana01 (approved):
It is CSS so I only looked briefly and looks fine as per code part

PR #917, opened 2026-06-25
Add sync-athena workflow

Hahihula:


alirana01 (approved):
It works good enough maybe we can mvoe the actions to the espressif namespace

Issue #921, opened 2026-06-29
[Bug Report] Issue with ESP-IDF Installation (EIM-888)


mjia948-boop:
## System Information
- **OS info**: OS: Windows 11 (26200) | Architecture: x86_64 | Kernel: 26200
- **App Version**: 0.15.0

## Description
Please describe the issue you encountered:

## Steps to Reproduce
插件全选，安装过程中一直让我登录极狐，https://jihulab.com/
我新用户，使用谷歌注册极狐后，找不到密码和账户，一直循环这个步骤。
我不知道极狐这个是不是有名，如果很有名气或者自家维护的，能不能不要搞得这么复杂啊
希望设置个自动登录，或者免登录的功能

## Logs
Please attach logs from the app logs folder.

Hahihula:
Thanks for the feedback, and sorry for the frustrating installation experience.

Unfortunately, in mainland China, or in any environment with unstable internet connectivity, the offline installation method is currently highly recommended, as it avoids many of the issues related to downloading components and accessing mirrors during the installation process. The documentation is available here:

https://docs.espressif.com/projects/idf-im-ui/en/latest/offline_installation.html

Regarding JiHuLab, we understand that the login flow can be confusing, especially for new users. We are currently migrating to our own Espressif-hosted mirror, which is already available at:

https://git.espressif.com.cn/

You can already use this mirror by setting the `idf_mirror` option in the installation configuration. More information can be found here:

https://docs.espressif.com/projects/idf-im-ui/en/latest/configuration.html

The good news is that this mirror is planned to become the default selection in the next release, which should eliminate the need for the JiHuLab login flow for most users in mainland China.

Thank you again for the report and for helping us improve the installation experience.

Issue #729, opened 2026-04-26
使用EIM安装时，需要提供jihulab.com的用户名和密码 (EIM-737)


leashi:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

<img width="1202" height="1032" alt="Image" src="https://github.com/user-attachments/assets/dbcea95b-3d83-4993-b97a-22344ac0604b" />

### Actual behavior (suspected bug)

使用EIM安装时，需要提供jihulab.com的用户名和密码

### Error logs or terminal output

```shell
使用EIM安装时，需要提供jihulab.com的用户名和密码
```

### Steps to reproduce the behavior

使用EIM安装时，需要提供jihulab.com的用户名和密码

### Project release version

v0.11.1

### System architecture

Intel/AMD 64-bit (modern PC, older Mac)

### Operating system

Windows

### Operating system version

windows 10, ubuntu24 上都存在

### Shell

ZSH

### Additional context

使用EIM安装时，需要提供jihulab.com的用户名和密码

gitlrj:
直接回车跳过应该就可以了吧?

bh1rio:
<img width="270" height="125" alt="Image" src="https://github.com/user-attachments/assets/de38c0f5-fa48-4ded-9855-582f6d4e8bb2" />
linux下，跳不过去

gitlrj:
https://jihulab.com/
极狐
试着注册一个? 然后输入账户密码看看??

win系统下也会老是弹出来,我注册了然后填入账户密码,
但是编译的时候又弹出来,我就直接不管了,直接治好我的强迫症了hhhhh

Hahihula:
You could respond with something like this:

Thanks for the report.

`jihulab.com` recently changed its access policy and now requires user registration/authentication for some repository access scenarios. Because of this, EIM installations using the JiHuLab mirror may prompt for a username and password.

We are currently working on providing our own dedicated China mirror infrastructure to avoid this dependency and improve the experience for users in mainland China.

For now, we recommend using the offline installation package, which is more suitable for mainland China environments:

Offline packages:
[...]

Issue #594, opened 2026-03-20
gitee credential required (EIM-622)


chinee0:
### Checklist

- [x] Checked the issue tracker for similar issues to ensure this is not a duplicate
- [x] Read the documentation to confirm the issue is not addressed there and your configuration is set correctly
- [x] Tested with the latest version to ensure the issue hasn't been fixed

### How often does this bug occurs?

always

### Expected behavior

easy mode and customized mode without selecting any mirror should not require gitee login

### Actual behavior (suspected bug)

<img width="2560" height="1350" alt="Image" src="https://github.com/user-attachments/assets/3fa90956-ca00-4a68-a74b-630963bcc580" /> 

### Error logs or terminal output

```shell
C:\Users\CVPC_002>eim-cli-windows-x64.exe wizard
2026-03-21 01:24:29 -  1 - 03 - INFO - Running IDF Installer Wizard...
2026-03-21 01:24:29 -  1 - 03 - INFO - All prerequisites are satisfied!
2026-03-21 01:24:29 -  1 - 03 - INFO - Running python sanity check
⠂ Doing something...                                                                                                    2026-03-21 01:24:29 -  1 - 03 - INFO - Using Scoop python3.exe: C:\Users\CVPC_002\scoop\shims\python3.exe
⠄ Doing something...                                                                                                      [PASS] App Execution Alias
  [PASS] Python Version
  [PASS] pip
  [PASS] venv
  [PASS] Standard Library (json, os, platform)
  [PASS] ctypes
  [PASS] SSL/HTTPS
2026-03-21 01:24:34 -  1 - 03 - INFO - Your Python meets the requirements
✔ Please select all of the target platforms (ESP chips) · all, esp32c3, esp32c6, esp32p4, esp32s3
✔ Please select the desired ESP-IDF version · v6.0
2026-03-21 01:24:49 -  1 - 03 - INFO - Starting mirror latency checks (2 candidates)...
2026-03-21 01:24:49 -  1 - 03 - INFO - Mirror score: https://github.com -> 356
2026-03-21 01:24:50 -  1 - 03 - INFO - Mirror score: https://jihulab.com/esp-mirror -> 824
✔ Select the source from which to download ESP-IDF · https://github.com (356 ms)
2026-03-21 01:24:52 -  1 - 03 - INFO - Starting mirror latency checks (3 candidates)...
2026-03-21 01:24:52 -  1 - 03 - INFO - Mirror score: https://github.com -> 505
[...]

PR #924, opened 2026-06-29
EIM-332: Replace Jihulab with our own mirror

Hahihula:
solves #729 #921 #594

### DIFF
--- a/tests/scripts/GUICustomInstall.test.js
+++ b/tests/scripts/GUICustomInstall.test.js
@@ -237,8 +237,8 @@ export function runGUICustomInstallTest({
       let githubMirror = await eimRunner.findByDataId(
         "idf-mirror-option-https://github.com",
       );
-      let jihulabMirror = await eimRunner.findByDataId(
-        "idf-mirror-option-https://jihulab.com/esp-mirror",
+      let espressifCnMirror = await eimRunner.findByDataId(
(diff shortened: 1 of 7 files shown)


brianignacio5:
Tested using China Unicom and works well.

Attaching logs for installation feedback.


[eim.log](https://github.com/user-attachments/files/29468858/eim.log)

tests/scripts/GUICustomInstall.test.js
```diff
@@ -237,36 +237,36 @@ export function runGUICustomInstallTest({
       let githubMirror = await eimRunner.findByDataId(
         "idf-mirror-option-https://github.com",
       );
-      let jihulabMirror = await eimRunner.findByDataId(
-        "idf-mirror-option-https://jihulab.com/esp-mirror",
+      let espressifCnMirror = await eimRunner.findByDataId(
```
alirana01, line 240: i guess better name would espressif s3 mirror as this can be better not now but aws i guess allows the resolution to the best mirror itself not sure if its implemented yet but that basically can be one service that can return the best mirror automatically can be done later just came into my mind while doing this but not critical maybe if you get time please also log a ticket for investigation so we can see if it helps simplify the rust side

alirana01 (approved):

alirana01:
Please just verify that tests are green else looks fine

PR #926, opened 2026-06-29
EIM-889: move git and python installer to s3

Hahihula:


src-tauri/src/lib/system_dependencies.rs
```diff
@@ -711,44 +711,118 @@ fn add_git_to_path(git_dir: &PathBuf) -> anyhow::Result<()> {
...
+        Ok(result) => {
+            debug!("Using S3 mirror for Git for Windows download");
+            return Ok(result);
+        }
+        Err(e) => {
+            debug!("S3 Git manifest unavailable ({}), falling back to GitHub API", e);
+        }
+    }
+
+    get_git_for_windows_url_from_github(arch_key).await
```
alirana01, line 753: Shouldn't we be doing the same for git as we do for python. Like the s3 can result a url and that can fail in that case we should try the url from github?

alirana01 (commented):
Generally looks fine can you please also verify that offline_installer_builder.rs has the same logic for python as well

Hahihula:
> Generally looks fine can you please also verify that offline_installer_builder.rs has the same logic for python as well

offline installer builder does not run in mainland china but in our pipeline.

alirana01 (approved):

PR #927, opened 2026-06-29
EIM-751: Part 2 updating the release orchestration workflows

alirana01:
**Refactor the release pipeline into three phases:**

1. release.yml creates a draft pre-release after all builds succeed
2. Maintainer reviews and publishes the draft
3. post-release.yml distributes to package repos with rollback on failure

### DIFF
--- a/.github/workflows/build-release.yaml
+++ b/.github/workflows/build-release.yaml
@@ -1,24 +1,14 @@
-name: Unified Build Workflow
+name: Release Build Workflow
 
 on:
-  schedule: [cron: "0 02 * * *"]
-  pull_request:
-    types:
-      - opened
-      - synchronize
-      - reopened
-      - ready_for_review
-    branches:
-      - master
-    paths-ignore:
-      - "dl_page/**"
-      - "docs/**"
-      - ".github/workflows/dl_page.yml"
-      - ".github/workflows/docs_build.yml"
-  release:
-    types:
-      - created
-  workflow_dispatch:
+  workflow_call:
+    inputs:
+      is_release:
+        type: boolean
+        default: false
+      version:
+        type: string
+        default: ''
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -80,7 +70,7 @@ jobs:
     permissions:
       contents: read
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix: ${{ fromJson(needs.setup.outputs.matrix) }}
     steps:
       - name: Checkout repository
@@ -136,7 +126,7 @@ jobs:
     permissions:
       contents: write
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         include:
           - os: ubuntu-latest
@@ -165,10 +155,9 @@ jobs:
           ZLIB_STATIC: 1
           LZMA_API_STATIC: 1
           APP_INSIGHTS_CONNECTION_STRING: ${{ secrets.APP_INSIGHTS_CONNECTION_STRING }}
-          # gix-specific environment variables
           GIX_HTTP_TIMEOUT: "60"
           GIX_MAX_CONNECTIONS: "1"
-          RUST_HTTP_PROTOCOL: "http/1.1"  # Disable HTTP/2 which can have issues on ARM
+          RUST_HTTP_PROTOCOL: "http/1.1"
         run: |
           cd src-tauri
           unset PKG_CONFIG_PATH && export ZLIB_STATIC=1 && export LZMA_API_STATIC=1 && export LIBZ_SYS_STATIC=1 && cargo build --release --no-default-features --features cli ${{ matrix.target && format('--target {0}', matrix.target) || '' }}
@@ -193,15 +182,12 @@ jobs:
 
           echo "=== Checking static linkage for: $BIN ==="
 
-          # ldd on a musl static binary exits non-zero and prints
-          # "not a dynamic executable" — we treat that as success.
           LDD_OUTPUT=$(ldd "$BIN" 2>&1 || true)
           echo "$LDD_OUTPUT"
 
           if echo "$LDD_OUTPUT" | grep -qE "not a dynamic executable|statically linked"; then
             echo "✓ Binary is statically linked."
           else
-            # Any resolved .so entry means a dynamic dependency leaked in.
             if echo "$LDD_OUTPUT" | grep -qE "\.so"; then
               echo "✗ Dynamic dependencies detected — binary is NOT fully static:" >&2
               echo "$LDD_OUTPUT" >&2
@@ -210,7 +196,6 @@ jobs:
             echo "✓ No dynamic libraries found — treating as static."
           fi
 
-          # Belt-and-suspenders: readelf / file as secondary checks
           if command -v readelf &>/dev/null; then
             INTERP=$(readelf -l "$BIN" 2>/dev/null | grep "INTERP" || true)
             if [ -n "$INTERP" ]; then
@@ -236,7 +221,6 @@ jobs:
           set -euo pipefail
           cd src-tauri
 
-          # Determine target output from cargo
           TRIPLE="${{ matrix.target }}"
           if [ -z "$TRIPLE" ]; then
             TRIPLE="x86_64-unknown-linux-musl"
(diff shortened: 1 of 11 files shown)


.github/workflows/post-release.yml
```diff
@@ -0,0 +1,186 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ap-east-1
+
+      - name: Fetch and upload release info
+        env:
+          DL_DISTRIBUTION_ID: ${{ secrets.DL_DISTRIBUTION_ID }}
+        run: |
+          curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            https://api.github.com/repos/${{ github.repository }}/releases/latest > eim_unified_release.json
```
Hahihula: Does `/release/latest` cover draft releases? Aren't we getting old info?

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,186 @@
...
+
+          DEFAULT_BRANCH=$(gh api repos/espressif/homebrew-eim --jq '.default_branch' 2>/dev/null || echo "main")
+          gh api repos/espressif/homebrew-eim/git/refs/heads/${DEFAULT_BRANCH} \
+            -X PATCH -f sha="$BACKUP_SHA" -F force=true || echo "::warning::Homebrew rollback failed"
+          echo "Homebrew rolled back"
+
+      - name: Rollback S3 APT repo
+        run: |
+          echo "Restoring APT repo from backup..."
+          aws s3 sync s3://espdldata/dl/eim/.backup/apt/ s3://espdldata/dl/eim/apt/ --acl public-read --delete --quiet || echo "::warning::APT rollback failed"
```
Hahihula: it seems to me like you are running `--delete` even in case the backup earlier failed. like if the backup repos fails, you are called using the `if: failure()` and than delete the data

.github/workflows/release.yml
```diff
@@ -0,0 +1,247 @@
...
+          echo "=== SHA256SUMS.txt ==="
+          cat SHA256SUMS.txt
+
+      - name: Create draft pre-release
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_PAT || secrets.GITHUB_TOKEN }}
+        run: |
+          gh release create "$TAG" \
+            --draft \
+            --prerelease \
```
Hahihula: --prerelease flag on draft requires undocumented manual step when publishing
 review flow requires maintainers to remember to uncheck "Set as a pre-release" before clicking Publish

.github/workflows/release.yml
```diff
@@ -0,0 +1,247 @@
...
+          for pkg in linux-x64 linux-aarch64; do
+            zip_file=$(find "artifacts/eim-gui-${pkg}-${VER}" -name "eim.zip" -type f 2>/dev/null | head -1)
+            if [[ -n "$zip_file" ]]; then
+              cp "$zip_file" "release_assets/eim-gui-${pkg}.zip"
+              echo "OK: eim-gui-${pkg}.zip"
+            fi
+          done
+
+          for pkg in macos-aarch64 macos-x64; do
+            zip_file=$(find "artifacts/eim-gui-${pkg}-${VER}" -name "*.zip" -type f 2>/dev/null | head -1)
```
Hahihula: this *.zip will probably cary more files than you thing

Hahihula (changes requested):
this pipeline is genuinely fragile in an asymmetric way — silent skips during build, hard failures during post-release: And all the silent fails in the first phase will cause hard failures i nthe second. hard to see, hard to debug, user has no visibility that the pipeline is already dead and will wait hour or more to relize so. For example in the case of npm hiccup which we encounter quite often in the piepeline.

.github/workflows/update-homebrew-release.yml
```diff
@@ -0,0 +1,285 @@
...
+          echo "version=${{ steps.tag.outputs.tag_name }}" >> $GITHUB_OUTPUT
+          echo "release_url=${{ github.event.release.html_url || format('https://github.com/{0}/releases/tag/{1}', github.repository, steps.tag.outputs.tag_name) }}" >> $GITHUB_OUTPUT
+
+      - name: Download and calculate checksums for CLI
+        id: cli-checksums
+        run: |
+          ASSETS=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" "${{ github.event.release.assets_url || format('https://api.github.com/repos/{0}/releases/tags/{1}', github.repository, steps.tag.outputs.tag_name) }}")
+
+          # macOS x64 CLI
+          macos_x64_url=$(echo "$ASSETS" | jq -r '.[] | select(.name == "eim-cli-macos-x64.zip") | .browser_download_url')
```
Hahihula: the jq will break on direct workflow_dispatch. `select(.name == "eim-cli-macos-x64.zip")` never matches, macos_x64_url is empty, and wget -q "" -O ... fails.

.github/workflows/update-windows-packages-release.yml
```diff
@@ -0,0 +1,168 @@
...
+          }
+          $guiJson | ConvertTo-Json -Depth 10 | Out-File -FilePath manifests-scoop/eim.json -Encoding UTF8
+
+          Write-Host "Generated Scoop manifests"
+
+      - name: Upload Scoop manifests to GitHub Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload ${{ inputs.version }} manifests-scoop/*.json --clobber
```
Hahihula: Every other gh call in the PR quotes "${{ inputs.version }}", why this does not?

Hahihula (changes requested):
great improvement with the exception of the jq looks great! GJ!

.github/workflows/build-release.yaml
```diff
@@ -0,0 +1,1691 @@
...
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Check PR merge to master in last 24 hours
```
Hahihula, line 29: why do you have this job here when you do not have schedule?

.github/workflows/build-release.yaml
```diff
@@ -0,0 +1,1691 @@
...
+        uses: actions/github-script@v7
+        with:
+          script: |
+            await github.rest.actions.cancelWorkflowRun({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: context.runId
+            });
+
+  build-offline-archives:
```
Hahihula, line 1077: Are you sure this will work without a change? You are not uploading the offline installer builder anymore in this workflow but still call workflow which expect it to be already uploaded.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,250 @@
...
+          aws-region: ap-east-1
+
+      - name: Backup repositories
+        id: backup
+        env:
+          GH_TOKEN: ${{ secrets.HOMEBREW_UPDATE_TOKEN || secrets.GITHUB_TOKEN }}
+        run: |
+          echo "tag=${TAG}" >> $GITHUB_OUTPUT
+
+          echo "=== Backup Homebrew HEAD ==="
```
Hahihula, line 71: looks like you are creating backups but never cleaning the up

.github/workflows/update-homebrew-release.yml
```diff
@@ -0,0 +1,309 @@
...
+
+      - name: Download and calculate checksums for CLI
+        id: cli-checksums
+        run: |
+          TAG_NAME="${{ steps.tag.outputs.tag_name }}"
+          RELEASE_JSON=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            "https://api.github.com/repos/${{ github.repository }}/releases/tags/${TAG_NAME}")
+          ASSETS=$(echo "$RELEASE_JSON" | jq '.assets')
+
+          download_asset() {
```
Hahihula, line 66: looks like you are declaring this function multiple times here. Please respect DRY

Hahihula (changes requested):

.github/workflows/build-release.yaml
alirana01: This is intentionally kept from the original `build.yaml` to minimize diff when merging back in Part 3 (replacing current workflows with this). The schedule check is effectively a no-op here since workflow_call never produces eventName == 'schedule' so it always sets should_run = true and falls through

.github/workflows/build-release.yaml
alirana01: The offline_installer_builder binary is still uploaded as a CI artifact (via actions/upload-artifact) we only removed the gh release upload steps. The build_offline_installer_archives workflow fetches it via gh run download from the same run ID, which reads CI artifacts, not release assets. So this works without changes

.github/workflows/post-release.yml
alirana01: This gets overwritten every release left it there intentionally so we may have chance to revert if we want manually later if we found anything critical after a while (maybe if we want a workflow for this can be added that will use the packages in backup to be reverted on repos if we publish and found the issues after a few hours)

Hahihula: It seems logical to either have a workflow for this or clean up the files after everything succeeds. I would probably prefer the latter as I expect the time between releases to get much longer as we are nearing the maintenance mode. But that's your call, either way, this seems to be an incomplete solution.

Hahihula (approved):

PR #931, opened 2026-07-01
Eim 432 add tools to fix command

Hahihula:
both CLI and GUI fix now allows user to specify list of tools so users can add more tools to existing installations.
Tools for v6 0 1
Tools for v6 0 1

### DIFF
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -742,10 +742,9 @@ pub async fn run_cli(cli: Cli) -> anyhow::Result<()> {
                 Err(err) => Err(anyhow::anyhow!(err)),
             }
         }
-        Commands::Fix { path } => {
-          let path_to_fix = if path.is_some() {
+        Commands::Fix { install_args } => {
+          let path_to_fix = if let Some(path) = install_args.path.clone() {
               // If a path is provided, fix the IDF installation at that path
-              let path = path.unwrap();
                if is_valid_idf_directory(&path) {
                 PathBuf::from(path)
                } else {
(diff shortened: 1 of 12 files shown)


src-tauri/src/cli/mod.rs
```diff
@@ -780,8 +779,12 @@ pub async fn run_cli(cli: Cli) -> anyhow::Result<()> {
             }
           };
           info!("{}", t!("fix.fixing", path = path_to_fix.display()));
-          // The fix logic is just instalation with use of existing repository
-          let settings = prepare_settings_for_fix_idf_installation(path_to_fix.clone(), config_path.as_ref()).await?;
+          // The fix logic is just instalation with use of existing repository.
+          // Start from the settings the installation was originally created with (so tools,
+          // features, target etc. are preserved), then let any CLI args the user explicitly
+          // passed to `fix` override those preserved values (and the defaults).
+          let mut settings = prepare_settings_for_fix_idf_installation(path_to_fix.clone(), config_path.as_ref()).await?;
```
alirana01, line 786: I am a bit cofnused on how the fix is working,
If I am not mistaken I think the `fix --config my.toml` will silently ignore the TOML as it may run the installation's recovered original settings or defaults plus any other explicit flags passed but nothing form `my.toml` is applied.

And since `--config` is flattened into `fix` its a valid flag and shows in `--help` so it looks _supported_ but the fix code path never loads it meanwhile the same `--config my.toml` will work on `install/wizard`

Just a bit confused on this maybe you can take a look and verify if this is intended behavior for it or maybe clarify the docs

alirana01 (approved):
Just a bit confused on the flag

PR #948, opened 2026-07-07
EIM-895: another attempt to fix the python dnf fedora dependency issue

Hahihula:
looks like the tauri builder ha toruble parsing the version boundary of the python package so we add it later after the tauri build the rpm


alirana01 (approved):
Good Luck 🤞

PR #998, opened 2026-07-29
Repo hardening: removing unwanted deps (basic machete run)

alirana01:
This is very initial run for the `cargo machete` I will dig bit more into that openssl dep to see how to get rid of that in later PR.


Hahihula (approved):
LGTM

PR #1019, opened 2026-08-05
Git bisect (docs and fix command improvements)

alirana01:
# Make `eim fix` cheap enough to run in a loop

## Summary

* `eim fix` no longer tears down and rebuilds the Python virtual environment on every run. It reuses the existing venv and runs pip without `--upgrade`, so only missing or unsatisfied packages are installed.
* If that pip run fails, the environment is deleted and retried **once** with `--upgrade`. A second failure is an error.
* `install` and `wizard` now follow the same path: reuse the venv first, nuke and retry on failure. There is no per-command switch controlling this behaviour.
* `fix` stops re-measuring mirror latency when the installation config was recovered. The recovered mirrors are preserved by setting `config_file` on the settings, which the existing latency-probing guard already checks.
* New guide: Using EIM with git bisect. `fix` is the command you run after changing the checked-out ESP-IDF revision yourself.

Together this takes a repeated `fix` on an unchanged installation from ~22.6 s to ~6 s of actual work (excluding a cold component-registry sync).

## What changed

**Uniform Python environment handling.** All three commands (`install`, `wizard`, `fix`) now follow the same path: try pip against the existing virtual environment without `--upgrade`; if that fails, delete the venv and retry once with `--upgrade`. The `WizardOptions` struct (which carried `try_existing_venv` and `preserve_mirrors`) is removed entirely, and `run_wizzard_run` goes back to taking only `Settings`. The `reinstall: bool` parameter on `install_python_env` is also removed, along with its early-return branch -- every caller now gets the retry path.

**Mirror preservation via `config_file`.** `prepare_settings_for_fix_idf_installation` now sets `settings.config_file` when the stored configuration is successfully recovered from the installation blob. The latency-probing guard in `select_single_mirror` (`else if needs_value && config.config_file.is_none()`) already skips probing when `config_file` is set, so recovered mirror values are preserved without a separate `preserve_mirrors` flag. No new fields are added to `Settings`.

**GUI call sites simplified.** `setup_tools` and `install_single_version` no longer take a `reinstall_python_env` parameter. All three GUI flows (install, fix, offline) call through the same uniform path.

**Documentation and locale updates.** The "Python environment" section in `cli_commands.md` now describes venv reuse as the behaviour for all commands. The `git_bisect.md` wording is adjusted accordingly. The `wizard.python.recreating_env` locale key is removed since the "recreate from scratch" log path no longer exists.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive guidance for using EIM with `git bisect`, including setup, automation, synchronization, cleanup, and troubleshooting.
  * Added documentation for derived version names and their use in environment paths and management commands.
  * Added links and FAQ guidance for bisecting ESP-IDF revisions.

* **Bug Fixes**
  * Improved Python environment installation recovery by reusing existing environments first, then recreating them with upgraded packages when installation fails.
  * Clarified configuration preservation, mirror handling, and recovery behavior across installation and fix workflows.


kolipakakondal:
@coderabbitai reivew

src-tauri/src/cli/cli_args.rs
```diff
@@ -141,6 +141,12 @@ pub enum Commands {
     Fix {
         #[command(flatten)]
         install_args: InstallArgs,
+
+        #[arg(
+            long,
+            help = "Whether to delete and rebuild the Python virtual environment. Defaults to false: the existing environment is reused and only the packages which are missing or no longer satisfy the constraints are installed. Set to true to get a clean environment and to upgrade packages to the newest version their constraint allows. Installation always rebuilds and is unaffected by this flag."
+        )]
+        recreate_py_env: Option<bool>,
```
Hahihula: every cli argument should have field in settings -> this approach will break in windows eim gui

src-tauri/src/cli/wizard.rs
```diff
@@ -335,7 +335,21 @@ async fn download_and_extract_tools(
...
 
-pub async fn run_wizzard_run(mut config: Settings) -> Result<(), String> {
+/// Per-run choices which are not part of the installation's configuration, so they are passed in
+/// rather than stored in `Settings` (and therefore never end up in the installation blob).
+pub struct WizardOptions {
+    /// Deletes the existing Python virtual environment and rebuilds it from scratch, upgrading
+    /// every package to the newest version its constraint allows. Installation passes `true`;
+    /// `fix` passes `false` unless the user asks for it, so that a repeated fix only installs the
+    /// packages which are missing or no longer satisfy the constraints.
+    pub recreate_py_env: bool,
```
Hahihula: this needs to be part of settings

src-tauri/src/cli/wizard.rs
```diff
@@ -335,7 +335,21 @@ async fn download_and_extract_tools(
...
+    /// The mirrors in `Settings` were recovered from an installation's stored configuration, so a
+    /// value which merely equals the built-in default is still a deliberate choice from the
+    /// original install and must not be re-derived by latency probing. Only `fix` sets this.
+    pub preserve_mirrors: bool,
```
Hahihula: Why do you thing this should be separated from settings?

Hahihula (changes requested):
the implementation of the cli commands is completly different from every other cli command and one of the switches isn't even exposed. Can you explain to me why did you choose this way and how to you plan to solve the issue with the unability to pass the cliccomands trough config file, which iws exactly how windows gui works?

src-tauri/src/cli/wizard.rs
alirana01: **It is separated because it is not a setting.** It does not mean "please keep using these mirror URLs" as a user preference. The mirror URLs themselves are already Settings fields (`mirror`, `idf_mirror`, `pypi_mirror`) and they **are** recovered from the blob. `preserve_mirrors` is the missing type information: "these values are recovered, so `is_default` lying about them must be ignored".

The distinction `is_default` cannot make:

| Situation | `idf_mirror` value | `is_default("idf_mirror")` | What `fix` should do |
| --- | --- | --- | --- |
| Fresh `Settings::default()`, no blob | `https://github.com` | `true` | Probe (this is `install`, or `fix` with no recoverable blob) |
| Blob from an install that chose GitHub | `https://github.com` | `true` | **Keep it. Do not probe.** |
| Blob from an install that chose git.espressif.com.cn | that URL | `false` | Keep it. Already worked before this PR. |
| User passed `--idf-mirror ...` on this `fix` | the override | depends | Keep the override (`apply_cli_overrides` already wins). |

Only row 2 is the bug. The signal that selects row 2 is `original_settings.is_some()` inside `prepare_settings_for_fix_idf_installation`, which is the only function that knows whether deserialization succeeded. That is why the plan threads a `bool` from there to `select_single_mirror` instead of storing anything new.

For GUI fix never calls `select_mirrors`. It uses `get_mirror_to_use` → `choose_mirror`, which already keeps the configured URL when `is_simple_installation` is false.

src-tauri/src/cli/cli_args.rs
alirana01: It is defined only on 'Fix` and documented as existing only on `fix`, and `install` / `wizard` do not accept it.
Since windows gui never spawns `eim fix` it never passes `--recreate-py-env`. I am a bit confused as there are commands that gui never uses nad has params like select rename remove they dont use that Settings convention. If we later want the fix to use this on gui we cna pass it when spawning fix

src-tauri/src/cli/wizard.rs
alirana01: If a user runs `eim fix --recreate-py-env true` once, the successful run writes Settings back into the blob. The next `eim fix` with no flags would recover `true` and wipe the venv again. The cheap path would only work until the first forced rebuild. Also adding a field breaks the bincode decode of every existing config fix would stop recovering tools/features/mirros for installs created by v0.18.0

src-tauri/src/cli/wizard.rs
Hahihula: ok this makes sense

src-tauri/src/cli/wizard.rs
```diff
@@ -335,7 +335,21 @@ async fn download_and_extract_tools(
...
 
-pub async fn run_wizzard_run(mut config: Settings) -> Result<(), String> {
+/// Per-run choices which are not part of the installation's configuration, so they are passed in
+/// rather than stored in `Settings` (and therefore never end up in the installation blob).
+pub struct WizardOptions {
+    /// Try pip against the existing virtual environment first (no `--upgrade`). Only `fix`
+    /// sets this; `install` / `wizard` leave it off so they still rebuild from scratch. If
+    /// that first attempt fails, `install_python_env` deletes the venv and retries once with
+    /// `--upgrade`.
+    pub try_existing_venv: bool,
```
Hahihula: i am sure we agreed that we do not want this switch that the workflow shoul be that we always try to reuse the env and if it fails tan recreate it

Hahihula (changes requested):
i am sure we agreed that we do not want this switch that the workflow shoul be that we always try to reuse the env and if it fails tan recreate it

src-tauri/src/cli/wizard.rs
alirana01: Its nor user facing anymore the flag is basically just to force that install and wizard always destroys existing venv which can be the case with install of a specific path and same version name

src-tauri/src/cli/mod.rs
```diff
@@ -814,7 +814,9 @@ pub async fn run_cli(cli: Cli) -> anyhow::Result<()> {
                 Err(err) => Err(anyhow::anyhow!(err)),
             }
         }
-        Commands::Fix { install_args } => {
+        Commands::Fix {
+            install_args,
+        } => {
```
Hahihula: Can you please revert unnecesary formating changes?

src-tauri/src/cli/prompts.rs
```diff
@@ -248,7 +248,6 @@ where
     FGet: Fn(&Settings) -> &Option<String>,
     FSet: Fn(&mut Settings, String),
 {
-    // Interactive by default when non_interactive is None
```
Hahihula, line 251: why did you removed the comment?

Hahihula (changes requested):

Hahihula (approved):
Gj. Thanks for the patience!

Issue #952, opened 2026-07-09
[Bug Report] Issue with ESP-IDF Installation (EIM-904)


duulga:
## System Information
- **OS info**: OS: Windows 11 (26100) | Architecture: x86_64 | Kernel: 26100
- **App Version**: 0.17.1

## Description
Please describe the issue you encountered:

## Steps to Reproduce
1. Execute GUI EIM v0.17.1
2. Start installation
3. Start Easy Installation
4. ESP-IDF version: v6.0.2 (3.09 GB)
5. Start installation

## Expected Behavior
After the download is finished, EIM should normally install ESP-IDF v6.0.2.
## Actual Behavior
Failed to Start Installation. Downloaded archive size mismatch for archive_v6.0.2_windows-x64.zst: expected 3312573311, got 3308825005
## Logs
Please attach logs from the app logs folder.

## Additional Information

[eim.log](https://github.com/user-attachments/files/29829220/eim.log)

[eim_gui.log](https://github.com/user-attachments/files/29829222/eim_gui.log)

JoaGamo:
I think this bug was introduced with #892 , I can reproduce the issue on Linux Ubuntu 26.04 LTS Kernel 7.0.0-29-generic

Workaround is going into expert/custom installation and enabling features step by step until installation is finished.

Hahihula:
hi @duulga,
this is actually caused by fault in our release process. if you run the install during the release, the offline instalelr archives are aleready uploaded but the json file with the size information is uploaded after all of them are there. 
Sorry for your inconvenience and i am working on the fix right now, but as workaround it should be ok just to rerun the installation and it should work.

PR #1036, opened 2026-08-11
EIM-904: Update offline installer archives publish to be atomic

Hahihula:
should solve #952 
seems to be working: https://github.com/espressif/idf-im-ui/actions/runs/31677640497

## Summary by CodeRabbit

- **Enhancements**
  - Offline installer archives are staged and verified before becoming available for download.
  - Archive content types and file sizes are checked before release metadata is published.
  - Build, promotion, and publication status provide improved visibility.

- **Reliability**
  - Release publication stops when archive promotion or verification fails.
  - Associated logs move alongside promoted archives when available.
  - Temporary staging files are cleaned up after each run.
  - Debug runs perform promotion and cleanup without publishing release metadata.

### DIFF
--- a/.github/workflows/build_offline_installer_archives.yaml
+++ b/.github/workflows/build_offline_installer_archives.yaml
@@ -69,6 +69,7 @@ jobs:
       version_count:    ${{ steps.get-versions.outputs.version_count }}
       should_purge:     ${{ steps.get-versions.outputs.should_purge }}
       debug:            ${{ steps.get-versions.outputs.debug }}
+      staging_prefix:   ${{ steps.get-versions.outputs.staging_prefix }}
       resolved_run_id:  ${{ steps.resolve-run-id.outputs.run_id }}
 
     steps:
@@ -138,6 +139,7 @@ jobs:
         run: |
           DEBUG="${{ inputs.debug }}"
           echo "debug=$DEBUG" >> $GITHUB_OUTPUT
+          echo "staging_prefix=release-staging/${{ github.run_id }}/" >> $GITHUB_OUTPUT
 
           # Purge logic
           if [ "$DEBUG" = "true" ]; then
@@ -174,6 +176,7 @@ jobs:
           DEBUG="${{ steps.get-versions.outputs.debug }}"
           PURGE="${{ steps.get-versions.outputs.should_purge }}"
           RUN_ID="${{ steps.resolve-run-id.outputs.run_id }}"
+          STAGING="${{ steps.get-versions.outputs.staging_prefix }}"
 
           echo "# 🏗️ Offline Installer Build Pipeline" >> $GITHUB_STEP_SUMMARY
           echo "" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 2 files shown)


.github/workflows/build_offline_installer_archives.yaml
```diff
@@ -380,6 +529,22 @@ jobs:
...
+        shell: bash
+        env:
+          STAGING_PREFIX: ${{ needs.get-versions.outputs.staging_prefix }}
+        run: |
+          set -euo pipefail
+          BUCKET="espdldata"
+          STAGED_DIR="dl/eim/${STAGING_PREFIX}"
+
+          echo "🧹 Cleaning up staging prefix s3://${BUCKET}/${STAGED_DIR}"
+          aws s3 rm "s3://${BUCKET}/${STAGED_DIR}" --recursive || true
```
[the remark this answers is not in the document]
alirana01: this is a small guard and can be added to avoid any mishaps in case of any weird problem to avoid recursive cleanups

.github/workflows/build_offline_installer_archives.yaml
```diff
@@ -245,7 +254,145 @@ jobs:
...
+                "${staged_uri}" \
+                "${canonical_uri}" \
+                --acl=public-read \
+                --content-type="application/zstd" \
+                --metadata-directive REPLACE 2>&1 | tee -a "$PROMOTE_LOG"; then
+              echo "::error::Failed to promote ${staged_uri} → ${canonical_uri}" | tee -a "$PROMOTE_LOG"
+              FAILED=$((FAILED+1))
+              continue
+            fi
+            PROMOTED=$((PROMOTED+1))
```
alirana01, line 322: the promotion is applied sequentially leading to same issue like 4 succeeded and rest are left the archives are updated but the json is not maybe if we can update json as each archive is updated would be better

.github/workflows/build_single_offline_installer_version.yml
```diff
@@ -204,84 +175,109 @@ jobs:
...
+          STAGED_KEY="${STAGING_PREFIX}${FILENAME}"
 
           SIZE=$([ "${{ runner.os }}" = "macOS" ] && stat -f %z "$archive" || stat -c %s "$archive")
 
-          echo "☁️ Uploading ${FILENAME} ($(numfmt --to=iec $SIZE)) to s3://espdldata/dl/eim/${S3_PREFIX}"
+          echo "☁️ Uploading ${FILENAME} ($(numfmt --to=iec $SIZE)) to s3://espdldata/dl/eim/${STAGED_KEY}"
 
           aws s3 cp --acl=public-read --content-type="application/zstd" \
-            "$archive" "s3://espdldata/dl/eim/${S3_PREFIX}${FILENAME}"
+            "$archive" "s3://espdldata/dl/eim/${STAGED_KEY}"
```
alirana01, line 213: shouldn't we be checking the exit code here as well like windows or make sure we `set -e` for exit

alirana01 (commented):
just a few small comments to take a look else its ci so things are bound to happen that will surprise later

.github/workflows/build_offline_installer_archives.yaml
Hahihula: these are already there, we are just moving them, this is closes to atomic we can get

alirana01 (approved):

PR #1047, opened 2026-08-19
offline installer tests should no longer run on failed build

Hahihula:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved command completion detection to reliably capture exit codes and avoid premature results.
  - Autotests now run only when at least one installer build succeeds, reducing misleading test failures.

- **Improvements**
  - Build dashboards now report successful and failed builds more clearly.
  - Added a notice when autotests are skipped because no build completed successfully.
  - Improved build dependency caching for more consistent installer builds.

proof this is working when launched with "wrong version" the tests are not launched blocking the pipeline
https://github.com/espressif/idf-im-ui/actions/runs/32263551463

### DIFF
--- a/.github/workflows/build_offline_installer_archives.yaml
+++ b/.github/workflows/build_offline_installer_archives.yaml
@@ -129,7 +129,8 @@ jobs:
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
           enable-cache: true
-          cache-dependency-glob: "uv.lock"
+          cache-dependency-glob: "src-tauri/Cargo.lock"
+          cache-suffix: ${{ inputs.idf_version }}
 
       # ── Compute versions & flags ───────────────────────────────────────────
       - name: Get IDF versions
@@ -553,7 +554,6 @@ jobs:
         with:
           pattern: build-info-*
           path: ./all-build-infos/
-        continue-on-error: true
 
       - name: Count successful builds
         id: count
--- a/.github/workflows/build_offline_installer_archives.yaml
+++ b/.github/workflows/build_offline_installer_archives.yaml
@@ -383,7 +383,7 @@ jobs:
   # ── 4. Final dashboard ─────────────────────────────────────────────────────
   dashboard:
     name: "📊 Pipeline Dashboard"
-    needs: [get-versions, build-version, update-json]
+    needs: [get-versions, build-version, update-json, count-successes]
     if: always()
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build_single_offline_installer_version.yml
```diff
@@ -115,12 +115,18 @@ jobs:
 
       - name: Install dependencies (macOS)
         if: runner.os == 'macOS'
+        env:
+          HOMEBREW_NO_REQUIRE_TAP_TRUST: "1"
```
alirana01, line 119: given the coderabbit comment i guess it makes sense since if the packages installed are already from a trusted tap also homebrew mentions it will be removed in later release but I guess nothing critical here

.github/workflows/build_single_offline_installer_version.yml
```diff
@@ -115,12 +115,18 @@ jobs:
...
         run: |
           brew update
           brew install gdk-pixbuf cairo pango libjpeg libpng zlib meson glib dbus-glib
 
       - name: Install UV
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          enable-cache: true
+          cache-dependency-glob: "src-tauri/Cargo.lock"
```
alirana01, line 128: confused on why we need it on cargo.lock file shouldn't this be on some python requirements

.github/workflows/build_offline_installer_archives.yaml
```diff
@@ -525,11 +533,66 @@ jobs:
...
+      success_count: ${{ steps.count.outputs.success_count }}
+      failure_count: ${{ steps.count.outputs.failure_count }}
+      any_success:   ${{ steps.count.outputs.any_success }}
+    steps:
+      - name: Download all build-infos
+        uses: actions/download-artifact@v8
+        with:
+          pattern: build-info-*
+          path: ./all-build-infos/
+        continue-on-error: true
```
alirana01: if this fails to download artifact which may have been built successfully due to github issue we may have some weird error information to diagnose in later steps

.github/workflows/build_offline_installer_archives.yaml
```diff
@@ -129,7 +129,8 @@ jobs:
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
           enable-cache: true
-          cache-dependency-glob: "uv.lock"
+          cache-dependency-glob: "src-tauri/Cargo.lock"
```
alirana01, line 132: confusing on having a uv with Cargo.lock cache dependency glob

alirana01 (dismissed):
Just some minor changes that are not critical but rather personal understanding of how the CI should be working

.github/workflows/build_single_offline_installer_version.yml
Hahihula: no they are not the aws is not trusted and it generates constant warnings in our CI

.github/workflows/build_single_offline_installer_version.yml
Hahihula: this is used as key for the cache, if cargo.locks changes, refresh the cache

alirana01 (approved):
LGTM!

PR #1066, opened 2026-08-26
EIM-878: telemetry update

Hahihula:
## Summary by CodeRabbit

* **New Features**
  * Added structured telemetry for CLI and GUI installation activity, including start, success, and failure outcomes.
  * Added privacy safeguards that redact personal information and categorize errors before telemetry is sent.
  * The global `--do-not-track` option now consistently disables telemetry for CLI commands and CLI-launched GUI sessions.

* **Documentation**
  * Expanded privacy and telemetry documentation to describe collected and excluded data, destinations, and disablement options.

### DIFF
--- a/src-tauri/src/gui/commands/utils_commands.rs
+++ b/src-tauri/src/gui/commands/utils_commands.rs
@@ -1,17 +1,20 @@
 use idf_im_lib::{self, ensure_path};
-use idf_im_lib::telemetry::{get_linux_os_name, track_event};
+use idf_im_lib::telemetry::{
+    self, get_linux_os_name, InstallMode, InstallOutcome, Interface,
+};
 use log::{error, info};
-use serde_json::{json,Value};
+use serde_json::{json, Value};
 use tauri_plugin_store::StoreExt;
 use std::fs;
 use std::{
     path::PathBuf,
     process::Command,
 };
-use tauri::AppHandle;
+use tauri::{AppHandle, Manager};
 use anyhow::{Result};
 use sysinfo::System;
 
+use crate::gui::app_state::AppState;
 use crate::gui::ui::send_message;
 
 const EIM_VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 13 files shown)


src-tauri/src/lib/telemetry.rs
```diff
@@ -16,53 +21,308 @@ static CLIENT: Lazy<Option<reqwest::Client>> = Lazy::new(|| {
...
+    error_kind: Option<ErrorKind>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error_message: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error_hash: Option<String>,
+    #[serde(flatten)]
+    extras: OutcomeExtras,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    subcommand: Option<String>,
+}
```
alirana01, line 271: can you verify if we are actually sennding the system information and eim version upstream?

src-tauri/src/lib/telemetry.rs
```diff
@@ -16,53 +21,308 @@ static CLIENT: Lazy<Option<reqwest::Client>> = Lazy::new(|| {
...
+            || msg.contains("prerequisite")
+            || msg.contains("dependency")
+        {
+            ErrorKind::DependencyMissing
+        } else if msg.contains("config") || msg.contains("invalid") || msg.contains("parse") {
+            ErrorKind::Configuration
+        } else {
+            ErrorKind::Unknown
+        }
+    }
```
alirana01, line 115: I am not sure if this is very good idea to check error messages i see some problems like python not found will go into the `UserCancelled` and basically the `ErrorKind::Python` never reaches if the first one is caught, since its a telemetry fix it is not causing any user side issues but the whole goal is to make sure we actually get proper telemetry I am not sure if this is ideal solution.

Another i believe this will also fail if the user has an other language selected.

I believe we need to have proper error classes so we can then simply use those enums here to figure out the kind when an error happens

src-tauri/src/gui/commands/utils_commands.rs
```diff
@@ -293,30 +296,114 @@ pub async fn install_drivers() -> Result<(), String> {
...
+                _ => None,
+            };
+            let extras = telemetry::OutcomeExtras {
+                feature_count,
+                tool_count,
+                target_count,
+                non_interactive,
+                used_existing_idf,
+            };
+            telemetry::track_install_outcome(&ctx, outcome, kind, error.as_ref(), extras);
```
alirana01: Can you also take a look here I am not sure how Vue passes the error information i think that erro_kind will be None and the function will fall through to None and possibly we falll into old trap of not proper telemetry again? please verify it maybe passing but I dont remember Vue working this way during an older PR experiment i was doing.

alirana01 (changes requested):
I think the general idea seems fine but I still think this needs some more work if we want a proper telemetry to reflect the actual issues this still has some quirks like for me the biggest one is the way the `telemetry.rs` tries to see `ErrorKind` this will possibly require a more bigger overhaul of how the EIM errors are handled

src-tauri/src/lib/telemetry.rs
Hahihula: i added eim version and system information

src-tauri/src/gui/commands/utils_commands.rs
```diff
@@ -293,30 +296,120 @@ pub async fn install_drivers() -> Result<(), String> {
...
+                .map_err(|_| "Failed to lock telemetry session".to_string())?;
+            *session = Some(ctx);
+        }
+        "install_finished" => {
+            let ctx = {
+                let mut session = app_state
+                    .telemetry_session
+                    .lock()
+                    .map_err(|_| "Failed to lock telemetry session".to_string())?;
+                session.take()
```
alirana01, line 358: not sure how this will behave in case of multiple versions selected by the user?

src-tauri/src/lib/telemetry.rs
alirana01: @Hahihula the ordering is definitely better maybe okay for initial telemetry update but it still has that issue of strings coming from the i18n and that can easily be a cn i believe we may need to refactor this in another pr for better telemetry,

src/components/OfflineInstaller.vue
```diff
@@ -775,8 +771,11 @@ export default {
       this.installed_versions = [];
       this.timeStarted = new Date();
 
-      try { // tracking should never fail installation
-        await invoke("track_event_command", { name: "GUI offline installation started" });
+      try {
+        await invoke("track_event_command", {
+          event: "install_started",
+          mode: "offline"
+        });
```
alirana01, line 778: maybe since versions are available here we can pass it

src-tauri/src/lib/telemetry.rs
```diff
@@ -16,53 +21,345 @@ static CLIENT: Lazy<Option<reqwest::Client>> = Lazy::new(|| {
...
+        _ => return,
+    };
+    properties.insert(
+        "appVersion".to_string(),
+        serde_json::Value::String(EIM_VERSION.to_string()),
+    );
+    properties.insert(
+        "systemInfo".to_string(),
+        serde_json::Value::String(get_system_info()),
+    );
```
alirana01, line 303: If these cna be added to `EventProps` we wont have to serialize twice minor stuff not that important

alirana01 (approved):
Thanks for the changes but still we need to create another tracking ticket to improve on this as the error parts is not ideal for telemetry filter but that will require much more overhaul of the project

